PR #9, opened 2018-01-19
Add storage interface

nikunjgit:


models/tag.go
```diff
@@ -0,0 +1,58 @@
+package models
```
benraskin92: Maybe add a couple unit tests for this package? Unless you were planning on doing that in another diff.

models/tag.go
```diff
@@ -0,0 +1,58 @@
...
+	typeToStr := map[MatchType]string{
+		MatchEqual:     "=",
+		MatchNotEqual:  "!=",
+		MatchRegexp:    "=~",
+		MatchNotRegexp: "!~",
+	}
+	if str, ok := typeToStr[m]; ok {
+		return str
+	}
+	panic("unknown match type")
```
benraskin92: Should we `panic` or return an error here?

nikunjgit: String() can't really return an error. Also, this shouldn't have any other value.

ts/series.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"time"
+)
+
+// A Series is the public interface to a block of timeseries values.  Each block has a start time,
+// a logical number of steps, and a step size indicating the number of milliseconds represented by each point.
+type Series struct {
+	name      string
```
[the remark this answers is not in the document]
nikunjgit: Added a method

storage/interface.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/m3db/m3coordinator/models"
+	"github.com/m3db/m3coordinator/ts"
+
+	xtime "github.com/m3db/m3x/time"
+)
+
+// Storage provides an interface for reading and writing to the tsdb
+type Storage interface {
+	Queryable
```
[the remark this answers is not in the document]
nikunjgit: done

PR #10, opened 2018-01-23
Add m3db client support

benraskin92:


services/m3coordinator/main/main.go
```diff
@@ -49,8 +52,31 @@ func main() {
...
+	m3dbClient, err := m3dbClientOpts.NewClient(client.ConfigurationParameters{})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to create m3db client, got error %v\n", err)
+		os.Exit(1)
+	}
+
+	session, err := m3dbClient.NewSession()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to create m3db client session, got error %v\n", err)
+		os.Exit(1)
```
nikunjgit: should we have a way to cleanup other things ? For example, you might need to call  m3dbClient.Close() ?

PR #13, opened 2018-01-26
Implement write endpoint

benraskin92:


services/m3coordinator/handler/write.go
```diff
@@ -0,0 +1,76 @@
...
+	var req prompb.WriteRequest
+	if err := proto.Unmarshal(reqBuf, &req); err != nil {
+		return nil, NewParseError(err, http.StatusBadRequest)
+	}
+
+	return &req, nil
+}
+
+func (h *PromWriteHandler) write(r *prompb.WriteRequest) error {
+	for _, t := range r.Timeseries {
```
nikunjgit: add TODO to parallelize

models/tag.go
```diff
@@ -81,12 +82,36 @@ func (m *Matcher) Matches(s string) bool {
 // Matchers is of matchers
 type Matchers []*Matcher
 
+// ToTags converts Matchers to Tags
```
nikunjgit: Maybe make it clear in the comments that this works only for exact matches

models/tag.go
```diff
@@ -81,12 +82,36 @@ func (m *Matcher) Matches(s string) bool {
...
+		if v.Type != MatchEqual {
+			return nil, fmt.Errorf("illegal match type, got %v, but expecting: %v", v.Type, MatchEqual)
+		}
+		tags[v.Name] = v.Value
+	}
+
+	return tags, nil
+}
+
 // ID returns a string representation of the matchers
```
nikunjgit: old comment ?

services/m3coordinator/handler/write.go
```diff
@@ -0,0 +1,76 @@
...
+		tagsList := storage.PromWriteTSToM3(t)
+		id := tagsList.ID()
+
+		for _, sample := range t.Samples {
+			timestamp := time.Unix(0, sample.Timestamp*int64(time.Millisecond))
```
nikunjgit: use PromTimestampToTime from convertor ?

services/m3coordinator/handler/write.go
```diff
@@ -0,0 +1,76 @@
...
+			if err := h.store.Write(id, timestamp, sample.Value, xtime.Millisecond, nil); err != nil {
```
nikunjgit: maybe change the store interface to have models.WriteQuery similar to read interface with models.ReadQuery ? Also, do you think we should allow the store interface to take multiple datapoints so that we can batch them to m3db later ?

storage/convertor.go
```diff
@@ -9,6 +9,23 @@ import (
 	"github.com/m3db/m3coordinator/ts"
 )
 
+// PromWriteTSToM3 converts a prometheus write query to an M3 one
+func PromWriteTSToM3(timeseries *prompb.TimeSeries) *models.Tags {
+	tagMatchers := PromLabelsToM3Tags(timeseries.Labels)
+
```
nikunjgit: extra line ?

services/m3coordinator/handler/write.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, NewParseError(err, http.StatusBadRequest)
+	}
+
+	return &req, nil
+}
+
+func (h *PromWriteHandler) write(r *prompb.WriteRequest) error {
+	for _, t := range r.Timeseries {
+		tagsList := storage.PromWriteTSToM3(t)
+		id := tagsList.ID()
```
nikunjgit: You probably don't want to get the id here. Different storages may define ID differently. So maybe just pass taglist to write and let the storage decide how to get the id

services/m3coordinator/handler/write.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return &req, nil
+}
+
+func (h *PromWriteHandler) write(ctx context.Context, r *prompb.WriteRequest) error {
+	for _, t := range r.Timeseries {
+		tagsList, datapoints := storage.PromWriteTSToM3(t)
+		writeQuery := createWriteQuery(tagsList, datapoints, nil)
+
```
nikunjgit: remove line

storage/convertor.go
```diff
@@ -9,14 +9,42 @@ import (
 	"github.com/m3db/m3coordinator/ts"
 )
 
+// PromWriteTSToM3 converts a prometheus write query to an M3 one
+func PromWriteTSToM3(timeseries *prompb.TimeSeries) (models.Tags, ts.Datapoints) {
```
nikunjgit: instead of doing this and then calling createWriteQuery() why not create a function PromWriteQueryToM3 which returns *WriteQuery ?

storage/local/storage.go
```diff
@@ -58,14 +56,27 @@ func (s *localStorage) Fetch(ctx context.Context, query *models.ReadQuery) (*sto
...
 	seriesList[0] = series
 	return &storage.FetchResult{
 		SeriesList: seriesList,
 	}, nil
 }
 
-func (s *localStorage) Write(tags models.Tags, t time.Time, value float64, unit xtime.Unit, annotation []byte) error {
+func (s *localStorage) Write(ctx context.Context, query *storage.WriteQuery) error {
+	id := query.Tags.ID()
+
```
nikunjgit: remove line

PR #17, opened 2018-02-06
Add execution engine and tracker for query states

nikunjgit:


storage/fanout/storage.go
```diff
@@ -17,10 +17,10 @@ func NewStorage(stores []storage.Storage, filter filter.Storage) storage.Storage
 	return &fanoutStorage{stores: stores, filter: filter}
 }
 
-func (s *fanoutStorage) Fetch(ctx context.Context, query *storage.ReadQuery) (*storage.FetchResult, error) {
+func (s *fanoutStorage) Fetch(ctx context.Context, query *storage.ReadQuery, options *storage.FetchOptions) (*storage.FetchResult, error) {
```
arnikola: Either `ReadQuery` -> `FetchQuery`
 or `FetchOptions` -> `ReadOptions`

storage/mock/storage.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func (s *mockStorage) Fetch(ctx context.Context, query *storage.ReadQuery, options *storage.FetchOptions) (*storage.FetchResult, error) {
+	time.Sleep(s.waitTime)
+	return s.result, nil
+}
+
+func (s *mockStorage) Write(ctx context.Context, query *storage.WriteQuery) error {
+	time.Sleep(s.waitTime)
+	return nil
```
arnikola: `return s.err`

storage/mock/storage.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/m3db/m3coordinator/storage"
+)
+
+type mockStorage struct {
+	waitTime time.Duration
+	result   *storage.FetchResult
```
arnikola: add `err error`

storage/mock/storage.go
```diff
@@ -0,0 +1,32 @@
...
+	return s.result, nil
+}
+
+func (s *mockStorage) Write(ctx context.Context, query *storage.WriteQuery) error {
+	time.Sleep(s.waitTime)
+	return nil
+}
+
+func (s *mockStorage) Type() storage.Type {
+	return storage.TypeMultiDC
```
arnikola: probably add a different type?

services/m3coordinator/handler/read_test.go
```diff
@@ -100,5 +99,58 @@ func TestQueryMatchMustBeEqual(t *testing.T) {
...
+	server := httptest.NewServer(promRead)
+	defer server.Close()
+
+	c := &http.Client{
+		Timeout: 1 * time.Millisecond,
+	}
+
+	_, err := c.Post(server.URL, "application/x-protobuf", generatePromReadBody(t))
 	assert.Error(t, err)
+	//	assert.NotNil(t, resp)
```
arnikola: should this be commented?

services/m3coordinator/handler/read.go
```diff
@@ -45,7 +45,7 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	result, err := h.read(r.Context(), w, req, params)
 	if err != nil {
-		logging.WithContext(r.Context()).Error("Read error", zap.Any("err", err))
```
arnikola: `logger := logging.WithContext(r.Context())` before this and use `logger`

errors/execution.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// ErrQueryAborted is an error returned when the query is aborted.
+	ErrQueryAborted = errors.New("query aborted")
+
+	// ErrQueryEngineShutdown is an error sent when the query cannot be
+	// created because the query engine was shutdown.
+	ErrQueryEngineShutdown = errors.New("query engine shutdown")
+
+	// ErrQueryTimeoutLimitExceeded is an error when a query hits the max time allowed to run.
+	ErrQueryTimeoutLimitExceeded = errors.New("query-timeout limit exceeded")
```
arnikola: `query timeout`
 also **the** query in comment

errors/execution.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// ErrMaxConcurrentQueriesLimitExceeded is an error when a query cannot be run
+// because the maximum number of queries has been reached.
+func ErrMaxConcurrentQueriesLimitExceeded(n, limit int) error {
+	return fmt.Errorf("max-concurrent-queries limit exceeded(%d, %d)", n, limit)
```
arnikola: `max-concurrent-queries`
also **the** query in comment

executor/engine_test.go
```diff
@@ -0,0 +1,36 @@
...
+	session.EXPECT().Fetch(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, fmt.Errorf("dummy"))
+
+	// Results is closed by execute
+	results := make(chan *storage.QueryResult)
+	closing := make(chan bool)
+
+	engine := NewEngine(local.NewStorage(session, "metrics", resolver.NewStaticResolver(policy.NewStoragePolicy(time.Second, xtime.Second, time.Hour*48))))
+	go engine.Execute(context.TODO(), &storage.ReadQuery{}, &EngineOptions{}, closing, results)
+	<- results
+	assert.Equal(t, len(engine.Tracker.queries), 1)
```
arnikola: Can we make 
```go
func (e *Engine) QueryLength() int {
 return len(e.tracker.queries)
}
```
then make `Tracker` private, since this seems to be the only place that uses it?

Otherwise, `assert.Len(t, engine.Queries(), 1)`

executor/engine.go
```diff
@@ -2,45 +2,15 @@ package executor
...
-	return fmt.Errorf("max-concurrent-queries limit exceeded(%d, %d)", n, limit)
-}
-
 // Engine executes a Query.
 type Engine struct {
 	// Used for tracking running queries.
-	tracker *Tracker
-	stats   *QueryStatistics
+	Tracker *Tracker
+	Stats   *QueryStatistics
```
arnikola: Does `stats` need to be exported?

services/m3coordinator/handler/read.go
arnikola: Also, error is implied and tagged; more descriptive message like "Unable to read handler context" maybe? Should also probably consolidate to "err" or "error" tag for errors

services/m3coordinator/handler/read.go
```diff
@@ -56,7 +56,7 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	data, err := proto.Marshal(resp)
 	if err != nil {
-		logging.WithContext(r.Context()).Error("Read error", zap.Any("err", err))
+		logging.WithContext(r.Context()).Error("unable to marshal read results to protobuf", zap.Any("err", err))
```
arnikola: unable -> Unable

services/m3coordinator/handler/read_test.go
```diff
@@ -100,5 +99,58 @@ func TestQueryMatchMustBeEqual(t *testing.T) {
...
+	mockResolver.EXPECT().Resolve(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Do(func(_, _, _, _ interface{}) {
+		time.Sleep(10 * time.Millisecond)
+	}).Return([]tsdb.FetchRequest{}, nil)
+
+	storage := local.NewStorage(session, "metrics", mockResolver)
+	engine := executor.NewEngine(storage)
+	promRead := &PromReadHandler{engine: engine}
+
+	server := httptest.NewServer(promRead)
+	defer server.Close()
```
arnikola: Up to here, this is duplicated setup code from `TestQueryKillOnClientDisconnect` except for the sleep duration; can pull out into a func?

services/m3coordinator/handler/read_test.go
```diff
@@ -100,5 +99,58 @@ func TestQueryMatchMustBeEqual(t *testing.T) {
...
+
+	req, _ := http.NewRequest("POST", server.URL, generatePromReadBody(t))
+	req.Header.Add("Content-Type", "application/x-protobuf")
+	req.Header.Add("timeout", "1ms")
+	resp, _ := http.DefaultClient.Do(req)
+	defer resp.Body.Close()
+	assert.NotNil(t, resp)
```
arnikola: This should probably be `require.NotNil` since `assert` keeps executing and next line may panic if this fails

services/m3coordinator/handler/read_test.go
```diff
@@ -100,5 +99,58 @@ func TestQueryMatchMustBeEqual(t *testing.T) {
...
+	assert.Equal(t, resp.StatusCode, 500, "Internal server error")
```
arnikola: the string bit here should be "Status code not 500" or something

services/m3coordinator/handler/read_test.go
nikunjgit: since we are forcing a timeout on the client and err is not nil then resp will always be nil

services/m3coordinator/handler/read.go
nikunjgit: usually error strings are not capitalized

executor/engine.go
nikunjgit: yeah, once we start tracking them, we will have to pass it around

PR #18, opened 2018-02-07
Add grpc server and client to enable remote fetch and write

arnikola:


services/m3coordinator/main/main.go
```diff
@@ -73,6 +76,18 @@ func main() {
...
 	handler.RegisterRoutes()
 
+	logger.Info("Creating gRPC server")
+	server := remote.CreateNewGrpcServer(ctx, storage)
+
+	go func() {
+		logger.Info("Starting gRPC server")
+		err = remote.StartNewGrpcServer(server, flags.rpcAddress)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to start gRPC server, got error %v\n", err)
```
nikunjgit: can you use logger ?

services/m3coordinator/main/main.go
```diff
@@ -106,6 +121,9 @@ func parseFlags() *m3config {
 	a.Flag("query.max-concurrency", "Maximum number of queries executed concurrently.").
 		Default("20").IntVar(&cfg.maxConcurrentQueries)
 
+	a.Flag("rpc.port", "Address to run rpc on.").
```
nikunjgit: better comment ?

storage/remote/storage.go
```diff
@@ -4,23 +4,28 @@ import (
...
 type remoteStorage struct {
-
+	client remote.Client
 }
 
 // NewStorage creates a new remote Storage instance.
-func NewStorage() storage.Storage {
-	return &remoteStorage{}
+func NewStorage(address string) (storage.Storage, error) {
+	c, err := remote.NewGrpcClient(address)
```
nikunjgit: can you make the client outside and pass to storage ? Makes testing easier

arnikola: Good call

services/m3coordinator/main/main.go
arnikola: There's a couple of places in main.go that use `fmt.*`; was copying that pattern, but will change the others too

services/m3coordinator/main/main.go
```diff
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"fmt"
+	"log"
```
nikunjgit: use our logging package and not this

ts/series.go
```diff
@@ -42,12 +42,15 @@ func (b *Series) StartTime() time.Time { return b.startTime }
...
-// Len returns the number of values in the time series.  Used for aggregation
+// Len returns the number of values in the time series. Used for aggregation
 func (b *Series) Len() int { return b.vals.Len() }
 
-// ValueAt returns the value at a given step.  Used for aggregation
+// ValueAt returns the value at a given step. Used for aggregation
 func (b *Series) ValueAt(i int) float64 { return b.vals.ValueAt(i) }
 
+// Values returns the values in the timeseries. Used for coversion
+func (b *Series) Values() Values { return b.vals }
```
nikunjgit: Not sure if we should have this. It makes the vals mutable

tsdb/remote/codecs.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/m3db/m3coordinator/storage"
+	"github.com/m3db/m3coordinator/ts"
+
+	"github.com/m3db/m3coordinator/models"
+	xtime "github.com/m3db/m3x/time"
+)
+
+func fromTime(t time.Time) int64 {
+	// Nanos -> Millis
+	return t.UnixNano() / 1000 / 1000
```
nikunjgit: t.UnixNano() / int64(time.Millisecond)

tsdb/remote/server.go
```diff
@@ -0,0 +1,74 @@
...
+
+func newServer(ctx context.Context, store storage.Storage) *grpcServer {
+	return &grpcServer{
+		ctx:     ctx,
+		storage: store,
+	}
+}
+
+// CreateNewGrpcServer creates server, given context local storage
+func CreateNewGrpcServer(ctx context.Context, store storage.Storage) *grpc.Server {
```
nikunjgit: Its a little strange to store ctx within grpcServer at creation since context is usually associated with a request.

tsdb/remote/server.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+
+	"github.com/m3db/m3coordinator/generated/proto/m3coordinator"
+
+	"github.com/m3db/m3coordinator/storage"
+	"google.golang.org/grpc"
```
nikunjgit: line before it ?

tsdb/remote/server.go
```diff
@@ -0,0 +1,74 @@
...
+	grpcServer := newServer(ctx, store)
+	rpc.RegisterQueryServer(server, grpcServer)
+	return server
+}
+
+// StartNewGrpcServer starts server on given address
+func StartNewGrpcServer(server *grpc.Server, address string) error {
+	lis, err := net.Listen("tcp", address)
+	if err != nil {
+		fmt.Println(err)
```
nikunjgit: should not print here and rely on the caller

arnikola: oops debugging print made it in

tsdb/remote/client.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"context"
+	"io"
+
+	"github.com/m3db/m3coordinator/generated/proto/m3coordinator"
+	"github.com/m3db/m3coordinator/ts"
+
+	"github.com/m3db/m3coordinator/storage"
+	"google.golang.org/grpc"
```
nikunjgit: separate this ?

tsdb/remote/client.go
```diff
@@ -0,0 +1,76 @@
+package remote
```
nikunjgit: we should probably figure out a uniform way for leaving blank lines

tsdb/remote/server.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Fetch reads from local storage
+func (s *grpcServer) Fetch(query *rpc.GrpcReadQuery, stream rpc.Query_FetchServer) error {
+	storeQuery, err := DecodeReadQuery(query)
+	if err != nil {
+		return err
+	}
+	result, err := s.storage.Fetch(s.ctx, storeQuery)
+	if err != nil {
+		return err
```
nikunjgit: probably want to log this error since its then send to remote ?

tsdb/remote/server.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return stream.Send(EncodeFetchResult(result))
+}
+
+// Write writes to local storage
+func (s *grpcServer) Write(stream rpc.Query_WriteServer) error {
```
nikunjgit: should we log the errors in this function as well ?

tsdb/remote/client.go
arnikola: I could write a formatting script that we could jam in our makefile

PR #19, opened 2018-02-12
Grpc

arnikola:


generated/proto/m3coordinator/query.proto
```diff
@@ -0,0 +1,60 @@
+syntax = "proto3";
+
+package rpc;
+
+service Query {
+// FetchQuery stream allows cancellation
+	rpc Fetch( FetchQuery) returns (stream FetchResult);
```
nikunjgit: separate query from options ?

services/m3coordinator/httpd/handler.go
```diff
@@ -63,10 +62,10 @@ func (h *Handler) registerProfileEndpoints() {
 func withResponseTimeLogging(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		startTime := time.Now()
-		// Attach a rqID with all logs so that its simple to trace the whole call stack
-		rqID := uuid.NewRandom()
-		rqCtx := logging.NewContext(r.Context(), zap.Stringer("rqID", rqID))
+
```
nikunjgit: extra line ?

nikunjgit: maybe we should only add a line after a code block ? (eg: if/else)

services/m3coordinator/main/main.go
```diff
@@ -34,46 +34,56 @@ var (
...
-		fmt.Fprintf(os.Stderr, "Unable to create m3db client, got error %v\n", err)
-		os.Exit(1)
+		logger.Fatal("Unable to create m3db client", zap.Any("error", err))
 	}
 
 	session, err := m3dbClient.NewSession()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Unable to create m3db client session, got error %v\n", err)
-		os.Exit(1)
+
```
nikunjgit: extra line ?

tsdb/remote/client.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/m3db/m3coordinator/generated/proto/m3coordinator"
+	"github.com/m3db/m3coordinator/storage"
+	"github.com/m3db/m3coordinator/ts"
+	"github.com/m3db/m3coordinator/util/logging"
+
+	"google.golang.org/grpc"
+)
+
+const (
+	maxRetries   = 10
```
nikunjgit: move these to config ?

tsdb/remote/client.go
```diff
@@ -0,0 +1,113 @@
...
+	storage.Querier
+	storage.Appender
+}
+
+type grpcClient struct {
+	client rpc.QueryClient
+}
+
+// NewGrpcClient creates grpc client
+func NewGrpcClient(address string) (Client, *grpc.ClientConn, error) {
```
nikunjgit: maybe instead of returning clientConn, define a Close() method on client itself

tsdb/remote/client.go
```diff
@@ -0,0 +1,113 @@
...
+	id := logging.ReadContextID(ctx)
+	var (
+		fetchClient rpc.Query_FetchClient
+		err         error
+	)
+	// Attempt to fetch, with incremental backoff if the server is busy
+	for delay, attempt := initialDelay, 0; ; attempt++ {
+		fetchClient, err = c.client.Fetch(ctx, EncodeFetchQuery(query, id))
+		if err != nil {
+			if strings.Contains(err.Error(), "the connection is unavailable") {
```
nikunjgit: is this a common pattern ? Seems strange to compare error string. Also maybe move backoff logic into a function in utils so that you can use it for both fetch and write ?

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
...
+		storage: store,
+	}
+}
+
+// CreateNewGrpcServer creates server, given context local storage
+func CreateNewGrpcServer(store storage.Storage) *grpc.Server {
+	server := grpc.NewServer()
+	grpcServer := newServer(store)
+	rpc.RegisterQueryServer(server, grpcServer)
+	logging.InitWithCores(nil)
```
nikunjgit: don't think you need to do that. Init should only be in main

tsdb/remote/codecs.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func encodeFetchOptions(queryID string) *rpc.FetchOptions {
+	return &rpc.FetchOptions{
+		Id: queryID,
+	}
+}
+
+// DecodeFetchQuery decodes rpc fetch query to read query
+func DecodeFetchQuery(query *rpc.FetchQuery) (*storage.FetchQuery, string, error) {
```
nikunjgit: once you split out FetchQuery with FetchOptions, this will probably change

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return server.Serve(lis)
+}
+
+// Fetch reads from local storage
+func (s *grpcServer) Fetch(query *rpc.FetchQuery, stream rpc.Query_FetchServer) error {
+	storeQuery, id, err := DecodeFetchQuery(query)
+	ctx := logging.NewContextWithID(stream.Context(), id)
+	logger := logging.WithContext(ctx)
+	defer logger.Sync()
```
nikunjgit: don't need this

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		logger.Error("Unable to decode fetch query", zap.Any("error", err))
+		return err
+	}
+
+	// Iterate while there are more results
+	for {
+		stream.Context()
```
nikunjgit: why do you need this ?

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
...
+
+		result, err := s.storage.Fetch(ctx, storeQuery, nil)
```
nikunjgit: I wonder if this should be parallelized.

arnikola: It could be potentially, depending on how we end up implementing multiple fetch responses. If the results are paginated and you need to call it again with a startIndex, parallelization won't work.

This is mostly here for future proofing and ensuring that multiple fetches get streamed correctly through the grpc calls, so it won't really do anything for real storage

tsdb/remote/codecs.go
```diff
@@ -0,0 +1,192 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/m3db/m3coordinator/generated/proto/m3coordinator"
+	"github.com/m3db/m3coordinator/storage"
+	"github.com/m3db/m3coordinator/ts"
+
+	"github.com/m3db/m3coordinator/models"
```
benraskin92: Move this import up one section

tsdb/remote/codecs.go
```diff
@@ -0,0 +1,192 @@
...
+	return storage.TimeToTimestamp(t)
+}
+
+func toTime(t int64) time.Time {
+	return storage.TimestampToTime(t)
+}
+
+// EncodeFetchResult encodes fetch result to rpc result
+func EncodeFetchResult(sResult *storage.FetchResult) *rpc.FetchResult {
+	series := make([]*rpc.Series, len(sResult.SeriesList))
```
benraskin92: Should this be `make([]*rpc.Series, 0, len(sResult.SeriesList))`?

tsdb/remote/codecs.go
```diff
@@ -0,0 +1,192 @@
...
+			Specification: result.Specification,
+			MillisPerStep: int32(result.MillisPerStep()),
+		}
+	}
+	return &rpc.FetchResult{Series: series}
+}
+
+// DecodeFetchResult decodes fetch results from a GRPC-compatible type.
+func DecodeFetchResult(ctx context.Context, rpcSeries []*rpc.Series) []*ts.Series {
+	tsSeries := make([]*ts.Series, len(rpcSeries))
```
benraskin92: Same here? `make([]*ts.Series, 0, len(rpcSeries))`

tsdb/remote/client.go
```diff
@@ -0,0 +1,101 @@
...
+func (c *grpcClient) Fetch(ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (*storage.FetchResult, error) {
+	id := logging.ReadContextID(ctx)
+	client, err := c.client.Fetch(ctx, EncodeFetchMessage(query, id))
+	if err != nil {
+		return nil, err
+	}
+
+	fetchClient, ok := client.(rpc.Query_FetchClient)
+	if !ok {
+		return nil, nil
```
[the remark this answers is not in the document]
arnikola: Good catch, the conversion was from a change I was trying but ended up reverting

tsdb/remote/client.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return &grpcClient{
+		client:     client,
+		connection: cc,
+	}, nil
+}
+
+// Fetch reads from remote client storage
+func (c *grpcClient) Fetch(ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (*storage.FetchResult, error) {
+	id := logging.ReadContextID(ctx)
```
nikunjgit: add a comment here about the id ?

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
+package remote
+
+import (
+	"io"
+	"net"
+
+	"github.com/m3db/m3coordinator/generated/proto/m3coordinator"
+	"github.com/m3db/m3coordinator/storage"
+	"github.com/m3db/m3coordinator/util/logging"
+	"go.uber.org/zap"
```
benraskin92: Move this to below import group

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
...
+
+func newServer(store storage.Storage) *grpcServer {
+	return &grpcServer{
+		storage: store,
+	}
+}
+
+// CreateNewGrpcServer creates server, given context local storage
+func CreateNewGrpcServer(store storage.Storage) *grpc.Server {
+	logging.InitWithCores(nil)
```
nikunjgit: don't think you need to do this

tsdb/remote/server.go
```diff
@@ -0,0 +1,102 @@
...
+
+	server := grpc.NewServer()
+	grpcServer := newServer(store)
+	rpc.RegisterQueryServer(server, grpcServer)
+
+	return server
+}
+
+// StartNewGrpcServer starts server on given address
+func StartNewGrpcServer(server *grpc.Server, address string, c chan<- struct{}) error {
```
nikunjgit: change c to be more informative ?

tsdb/remote/server.go
arnikola: Weird, thought I'd gotten rid of this already, must have reverted accidentally

services/m3coordinator/main/main.go
```diff
@@ -34,60 +34,70 @@ var (
...
 	handler, err := httpd.NewHandler(storage, executor.NewEngine(storage))
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Unable to set up handlers, got error %v\n", err)
-		os.Exit(1)
+		logger.Fatal("unable to set up handlers", zap.Any("error", err))
 	}
 	handler.RegisterRoutes()
 
-	logger.Info("Starting server", zap.String("address", flags.listenAddress))
+	logger.Info("creating gRPC server")
```
nikunjgit: extract the block to start a grpc server into a new func and also put it behind a flag since most people would start out by not wanting remote storage

tsdb/remote/client.go
```diff
@@ -0,0 +1,104 @@
...
+	client     rpc.QueryClient
+	connection *grpc.ClientConn
+}
+
+// NewGrpcClient creates grpc client
+func NewGrpcClient(addresses []string) (Client, error) {
+	if len(addresses) == 0 {
+		return nil, errors.ErrNoClientAddresses
+	}
+	resolver := newStaticResolver(addresses)
```
nikunjgit: figure out dynamic topology here ?

tsdb/remote/server.go
```diff
@@ -0,0 +1,100 @@
...
+func CreateNewGrpcServer(store storage.Storage) *grpc.Server {
+	server := grpc.NewServer()
+	grpcServer := newServer(store)
+	rpc.RegisterQueryServer(server, grpcServer)
+
+	return server
+}
+
+// StartNewGrpcServer starts server on given address, then notifies channel
+func StartNewGrpcServer(server *grpc.Server, address string, waitForStart chan<- struct{}) error {
```
nikunjgit: any reason not to make waitForStart chan bool ?/

tsdb/remote/client.go
arnikola: I think might be better in remote/storage, then pass a resolver down into client.

Either way, probably better to split out dynamic topology to a new pr

tsdb/remote/server.go
arnikola: struct{} has 0 size, and we're only using the channel for notifications

arnikola: struct{} has 0 size, and we're only using the channel for notifications

services/m3coordinator/main/main.go
arnikola: Sgtm

tsdb/remote/server.go
arnikola: That being said, may have fallen prey to the "this is how you do it in theory" approach that textbooks take :p

services/m3coordinator/main/main.go
```diff
@@ -15,8 +14,11 @@ import (
 	"github.com/m3db/m3coordinator/policy/resolver"
 	"github.com/m3db/m3coordinator/services/m3coordinator/config"
 	"github.com/m3db/m3coordinator/services/m3coordinator/httpd"
+	"github.com/m3db/m3coordinator/storage"
 	"github.com/m3db/m3coordinator/storage/local"
+	"github.com/m3db/m3coordinator/tsdb/remote"
 	"github.com/m3db/m3coordinator/util/logging"
+	"google.golang.org/grpc"
```
benraskin92: nit: add to end as a separate import group

tsdb/remote/codecs_test.go
```diff
@@ -0,0 +1,225 @@
+package remote
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	xtime "github.com/m3db/m3x/time"
```
benraskin92: nit: put this import below `m3coordinator` ones

tsdb/remote/server.go
```diff
@@ -0,0 +1,100 @@
...
+	logger := logging.WithContext(ctx)
+
+	if err != nil {
+		logger.Error("unable to decode fetch query", zap.Any("error", err))
+		return err
+	}
+
+	// Iterate while there are more results
+	for {
+		result, err := s.storage.Fetch(ctx, storeQuery, nil)
```
benraskin92: Not sure if this is necessary, but we could create a worker pool and use that. this way we won't be creating a new `result` object each time?

tsdb/remote/server_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"io"
+	"sync"
+	"testing"
+	"time"
+
+	m3err "github.com/m3db/m3coordinator/errors"
+	"github.com/m3db/m3coordinator/models"
+	"github.com/m3db/m3coordinator/storage"
+	"github.com/m3db/m3coordinator/ts"
+	"github.com/m3db/m3coordinator/util/logging"
```
benraskin92: nit: fix import ordering

tsdb/remote/server.go
arnikola: This is kinda future proofing, since we return the entire set of results rather than smaller bits at the moment. This code is assuming pagination, so parallel execution wouldn't really be possible. Right now, it will just run through once and finish, outside of tests specifically for this.

Also we have to create a new `result` each time since we're streaming them back to the remote `client` as they come back from the local storage.

services/m3coordinator/main/main.go
```diff
@@ -34,60 +36,78 @@ var (
...
 	maxConcurrentQueries int
 	queryTimeout         time.Duration
 }
 
 func main() {
 	rand.Seed(time.Now().UnixNano())
 
-	flags := parseFlags()
+	logging.InitWithCores(nil)
+	ctx := context.TODO()
```
nikunjgit: actually use context.Background(). Just read about it :)

tsdb/remote/client.go
nikunjgit: makes sense

tsdb/remote/server.go
nikunjgit: haha I didn't know about that :)

tsdb/remote/server.go
nikunjgit: maybe add a TODO and we can look at it during our perf improvements

PR #21, opened 2018-02-14
Update README and example configs

benraskin92:


benchmark/configs/benchmark.yml
```diff
@@ -5,12 +5,12 @@ client:
       topology:
           shards: 64
           hosts:
-            - host: localhost1
+            - host: m3db_server_1
+              listenAddress: "0.0.0.0:9000"
+            - host: m3db_server_2
+              listenAddress: "0.0.0.0:9000"
```
arnikola: Should these be 9010 and 9020?

README.md
```diff
@@ -52,6 +52,36 @@ Run Prometheus Docker image:
 
     $ docker run -p 9090:9090 -v $GOPATH/src/github.com/m3db/m3coordinator/docker/prometheus-mac.yml:/etc/prometheus/prometheus.yml quay.io/prometheus/prometheus
 
+### Running on GCP
+
+Setup GCP for m3db:
+
+    1. See m3db repo (m3db/example/README.md)
```
arnikola: Can't find where this is

benraskin92: Yeah, it's not in master yet, let me link to the PR.

benchmark/configs/benchmark.yml
benraskin92: These ports are just what the user configures m3db to listen on. 9000 is the default so figured we should leave it set to those. I only had 9020 and 9030 cause i was running all of this on one host.

benraskin92: These ports are just what the user configures m3db to listen on. 9000 is the default so figured we should leave it set to those. I only had 9020 and 9030 cause i was running all of this on one host.

PR #24, opened 2018-02-23
Improve write path perf

benraskin92:


errors/execution.go
```diff
@@ -9,6 +9,9 @@ var (
 	// ErrInvalidQuery is returned when executing an unknown query type.
 	ErrInvalidQuery = errors.New("invalid query")
 
+	// ErrNilQuery is returned when trying to write a nil query
+	ErrNilQuery = errors.New("write query must not be nil")
```
nikunjgit: ErrNilWriteQuery?

storage/local/storage.go
```diff
@@ -87,16 +89,52 @@ func (s *localStorage) Write(ctx context.Context, query *storage.WriteQuery) err
 	default:
 	}
 
+	if query == nil {
```
nikunjgit: do we really need this ? I think upstream should ensure this function cant get nil

storage/local/storage.go
```diff
@@ -87,16 +89,52 @@ func (s *localStorage) Write(ctx context.Context, query *storage.WriteQuery) err
...
+	writeRequestCommon := writeRequestCommon{
+		store:      s,
+		annotation: query.Annotation,
+		unit:       query.Unit,
+		id:         id,
+	}
+
+	requests := make([]execution.Request, len(query.Datapoints))
+	for idx, datapoint := range query.Datapoints {
+		requests[idx] = newWriteRequest(&writeRequestCommon, datapoint.Timestamp, datapoint.Value)
```
nikunjgit: isn't this a bug if all of them share writeRequestCommon and update it in parallel ?

storage/local/storage.go
```diff
@@ -87,16 +89,52 @@ func (s *localStorage) Write(ctx context.Context, query *storage.WriteQuery) err
...
+	return execution.ExecuteParallel(ctx, requests)
 }
 
 func (s *localStorage) Type() storage.Type {
 	return storage.TypeLocalDC
 }
+
+func (w *writeRequest) Process(ctx context.Context) error {
+	common := w.writeRequestCommon
+	store := common.store
```
nikunjgit: why separate only store out ?

services/m3coordinator/handler/write.go
```diff
@@ -58,11 +59,25 @@ func (h *PromWriteHandler) parseRequest(r *http.Request) (*prompb.WriteRequest,
...
+func (w *localWriteRequest) Process(ctx context.Context) error {
+	return w.store.Write(ctx, w.writeQuery)
+}
+
+type localWriteRequest struct {
+	store      storage.Storage
+	writeQuery *storage.WriteQuery
+}
+
+func newLocalWriteRequest(writeQuery *storage.WriteQuery, store storage.Storage) execution.Request {
```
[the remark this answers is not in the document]
benraskin92: I think it reads a little better

storage/local/storage.go
benraskin92: Yeah I wanted this in since one of our test checks for nil metrics

storage/local/storage.go
benraskin92: Yeah- store is used for the `session` and the `namespace`, while common is used for `id`, `unit`, and `annotation`.

PR #26, opened 2018-02-27
Arnikola/coordinator benchmarker

arnikola:
Both read and write benchmarkers for m3coordinator client.


benchmark/common/parse_json.go
```diff
@@ -40,6 +40,52 @@ type M3Metric struct {
...
+			go unmarshalMetrics(dataChannel, metricChannel, wg)
+		}
+		go func() {
+			for metric := range metricChannel {
+				f(metric)
+			}
+		}()
+	}
+	cleanup := func() {
+		wg.Wait()
```
nikunjgit: maybe workFunction should have wg.Wait() instead of cleanup

nikunjgit: Oh I see, I'd suggest that you get rid of cleanup and use a done chan.

benchmark/common/parse_json.go
```diff
@@ -40,6 +40,52 @@ type M3Metric struct {
...
+
+// ConvertToProm parses the json file that is generated from InfluxDB's bulk_data_gen tool into Prom format
+func ConvertToProm(fileName string, workers int, batchSize int, f func(*bytes.Reader)) {
+	metricChannel := make(chan *bytes.Reader, MetricsLen)
+	dataChannel := make(chan []byte, MetricsLen)
+	wg := new(sync.WaitGroup)
+	workFunction := func() {
+		for w := 0; w < workers; w++ {
+			wg.Add(1)
+			go marshalTsdbToProm(dataChannel, metricChannel, batchSize, wg)
```
nikunjgit: maybe create an anonymous func here and call wg.Done() after marshalTsdbToProm instead of doing that within marshalTsdbToProm.

benchmark/common/parse_json.go
```diff
@@ -112,3 +146,62 @@ func id(lowerCaseTags map[string]string, name string) string {
 
 	return buffer.String()
 }
+
+func marshalTsdbToProm(dataChannel <-chan []byte, metricChannel chan<- *bytes.Reader, batchSize int, wg *sync.WaitGroup) {
+	defer wg.Done()
+	timeseries := make([]*prompb.TimeSeries, batchSize)
+	idx := 0
+	for data := range dataChannel {
+		if len(data) != 0 {
```
nikunjgit: flip it to reduce nesting  if len(data) == 0 { continue}

benchmark/common/parse_json.go
```diff
@@ -112,3 +146,62 @@ func id(lowerCaseTags map[string]string, name string) string {
...
+			var m Metrics
+			if err := json.Unmarshal(data, &m); err != nil {
```
nikunjgit: maybe refactor the unmarshal part into another func ?

benchmark/write/write_benchmark.go
```diff
@@ -44,16 +45,29 @@ func init() {
 	flag.StringVar(&benchmarkers, "benchmarkers", "localhost:8888", "Comma separated host:ports addresses of benchmarkers to coordinate")
 	flag.BoolVar(&memprofile, "memprofile", false, "Enable memory profile")
 	flag.BoolVar(&cpuprofile, "cpuprofile", false, "Enable cpu profile")
+	flag.StringVar(&coordinatorAddress, "coordinatorAddress", "http://localhost:7201/api/v1/prom/write", "Address coordinator is listening on")
```
nikunjgit: call it write endpoint ?

policy/filter/storage.go
```diff
@@ -9,3 +9,8 @@ type Storage func(query storage.Query, store storage.Storage) bool
 func LocalOnly(query storage.Query, store storage.Storage) bool {
 	return store.Type() == storage.TypeLocalDC
 }
+
+// AllowAll does not filter any storages
+func AllowAll(_ storage.Query, _ storage.Storage) bool {
```
nikunjgit: maybe create a Static filter which takes in a value as a parameter and always returns that value. That can probably be useful in testing as well

services/m3coordinator/handler/read.go
```diff
@@ -39,11 +39,11 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	req, rErr := h.parseRequest(r)
+
 	if rErr != nil {
 		Error(w, rErr.Error(), rErr.Code())
 		return
 	}
-
```
nikunjgit: usually we leave a line after a code block ?

tsdb/remote/client.go
```diff
@@ -90,11 +90,15 @@ func (c *grpcClient) Write(ctx context.Context, query *storage.WriteQuery) error
 	id := logging.ReadContextID(ctx)
 	rpcQuery := EncodeWriteMessage(query, id)
 	err = writeClient.Send(rpcQuery)
-	if err != nil {
+
+	if err != nil && err != io.EOF {
 		return err
 	}
 
 	_, err = writeClient.CloseAndRecv()
```
nikunjgit: is this good to call if err == io.EOF ?

benchmark/write/write_benchmark.go
```diff
@@ -73,40 +87,102 @@ func main() {
...
+
+func genericBenchmarker(workerFunction func(), appendReadCount func() int, cleanup func()) {
 	if cpuprofile {
 		p := profile.Start(profile.CPUProfile)
 		defer p.Stop()
-	}
-
-	if memprofile {
+		if memprofile {
+			fmt.Println("cannot have both cpu and mem profile active at once; defaulting to cpu profiling")
```
nikunjgit: oh i didn't know that

benchmark/write/write_benchmark.go
```diff
@@ -120,40 +196,34 @@ func main() {
...
 }
 
-func addMetricsToChan(ch chan *common.M3Metric, wq []*common.M3Metric) int {
-	var items int
-	for _, query := range wq {
-		ch <- query
-		items++
+func writeToCoordinator(ch <-chan *bytes.Reader) {
+	for query := range ch {
+		if r, err := http.Post("http://localhost:7201/api/v1/prom/write", "", query); err != nil {
```
nikunjgit: use coordinatorAddress  ?

benchmark/common/parse_json.go
arnikola: I was meaning to do something like that and not pass around the `wg` but somehow forgot I could do it lol

tsdb/remote/client.go
arnikola: Yeah, EOF is an expected error from `.Send()` once it's done; I was assuming that `.CloseAndRecv()` returns `nil` if it succeeds, but it also returns an EOF

policy/filter/storage.go
arnikola: I think in this case where there's only 2 possible values for the parameter it might actually be a bit clearer if we added this and `AllowNone` which would return false

benchmark/common/http_utils.go
```diff
@@ -0,0 +1,19 @@
+package common
```
nikunjgit: the file name is a little too generic.

benchmark/common/parse_json.go
```diff
@@ -112,3 +146,66 @@ func id(lowerCaseTags map[string]string, name string) string {
...
+func marshalTsdbToProm(dataChannel <-chan []byte, metricChannel chan<- *bytes.Reader, batchSize int) {
+	timeseries := make([]*prompb.TimeSeries, batchSize)
+	idx := 0
+	for data := range dataChannel {
+		if len(data) == 0 {
+			continue
+		}
+		var m Metrics
+		if err := json.Unmarshal(data, &m); err != nil {
+			panic(err)
```
nikunjgit: we should probably figure out if we should panic on a single error or let it run. I think we do it differently at different places

benchmark/read/read_benchmark.go
```diff
@@ -0,0 +1,201 @@
...
+		matchers[i] = &prompb.LabelMatcher{
+			Type:  prompb.LabelMatcher_EQ,
+			Name:  "eq",
+			Value: id,
+		}
+	}
+	return matchers
+}
+
+func generatePromReadRequest(start, end time.Time) *prompb.ReadRequest {
```
nikunjgit: can you separate out request generation into its own file ? I'd assume we would work on making that better in the future instead of using a static request

policy/filter/storage.go
nikunjgit: sounds good

benchmark/common/parse_json.go
arnikola: I followed the existing conventions we had here; can get rid of the panics if preferred.

benchmark/read/read_benchmark.go
arnikola: It's kind of ingrained with getUniqueIds at the moment which is pretty awkward to make public or extract cleanly unfortunately

benchmark/common/parse_json.go
```diff
@@ -40,56 +40,89 @@ type M3Metric struct {
...
-			metricChannel <- &M3Metric{ID: id(m.Tags, m.Name), Time: storage.TimestampToTime(m.Time), Value: m.Value}
+		if len(data) == 0 {
+			continue
+		}
+		var m Metrics
+		if err := json.Unmarshal(data, &m); err != nil {
+			log.Fatalf("failed to unmarshal metrics, got error: %v\n", err)
 		}
+
+		metricChannel <- &M3Metric{ID: id(m.Tags, m.Name), Time: storage.TimestampToTime(m.Time), Value: m.Value}
```
benraskin92: Do we want to add `Tags` and test via the index write path? i.e.
```
type M3Metric struct {
	ID    ident.ID
	Tags  ident.Tags
	Time  time.Time
	Value float64
}
```

benchmark/common/parse_json.go
```diff
@@ -112,3 +145,66 @@ func id(lowerCaseTags map[string]string, name string) string {
...
+func metricsToPromTS(m Metrics) *prompb.TimeSeries {
+	labels := metricsTagsToLabels(m.Tags)
+	samples := metricsPointsToSamples(m.Value, m.Time)
+	return &prompb.TimeSeries{
+		Labels:  labels,
+		Samples: samples,
+	}
+}
+
+func marshalTsdbToProm(dataChannel <-chan []byte, metricChannel chan<- *bytes.Reader, batchSize int) {
```
benraskin92: nit: rename to `marshalTSDBToProm `

benchmark/common/parse_json.go
```diff
@@ -112,3 +145,66 @@ func id(lowerCaseTags map[string]string, name string) string {
...
+	req := &prompb.WriteRequest{
+		Timeseries: ts,
+	}
+	data, _ := proto.Marshal(req)
+	compressed := snappy.Encode(nil, data)
+	b := bytes.NewReader(compressed)
+	return b
+}
+
+func metricsTagsToLabels(tags map[string]string) []*prompb.Label {
```
benraskin92: I think we might have this in the `storage` package (i.e. `TagsToPromLabels`)

benchmark/read/read_benchmark.go
```diff
@@ -0,0 +1,201 @@
...
+import (
+	"bytes"
+	"flag"
+	"io"
+	"log"
+	"time"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	"github.com/m3db/m3coordinator/benchmark/common"
```
benraskin92: nit: import out of order

cc @nikunjgit

benchmark/read/read_benchmark.go
```diff
@@ -0,0 +1,201 @@
...
+	start = start.UTC()
+	end, err := time.Parse(time.RFC3339, timestampEndStr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	end = end.UTC()
+	// Split on coord vs m3db
+
+	if coordinator {
+		log.Println("Benchmarking on m3coordinator...")
```
benraskin92: Two things:

1. Do we want to stick with the `logger` package in `github.com/m3db/m3coordinator/util/logging` to be consistent?
2. Perhaps make it more clear what is being benchmarked?

benchmark/read/read_benchmark.go
```diff
@@ -0,0 +1,201 @@
...
+		benchmarkCoordinator(start, end)
+	} else {
+		log.Println("Benchmarking on m3db...")
```
benraskin92: Same here

benchmark/read/read_benchmark.go
```diff
@@ -0,0 +1,201 @@
...
+	})
+	ids := make([]string, 0, len(idMap))
+	for k := range idMap {
+		ids = append(ids, k)
+	}
+	return ids
+}
+
+func genericBenchmarker(fetch func(), count countFunc) {
+	start := time.Now()
```
benraskin92: nit: log when we start and end?

benchmark/write/benchmarker.go
```diff
@@ -146,7 +145,7 @@ func (b *benchmarker) waitForBenchmarkers() {
 		}()
 	}
 
-	fmt.Printf("all ready, now synchronizing to nearest 10s...\n")
+	fmt.Println("all ready, now synchronizing to nearest 10s...\n")
```
benraskin92: You can remove the `\n` since `fmt.Println` adds a newline at the end (`fmt.Printf` does not)

benchmark/write/write_benchmark.go
```diff
@@ -44,16 +44,31 @@ func init() {
...
 	flag.BoolVar(&memprofile, "memprofile", false, "Enable memory profile")
 	flag.BoolVar(&cpuprofile, "cpuprofile", false, "Enable cpu profile")
+	flag.StringVar(&writeEndpoint, "writeEndpoint", "http://localhost:7201/api/v1/prom/write", "Write endpoint for m3coordinator")
+	flag.BoolVar(&coordinator, "coordinator", false, "Benchmark through coordinator rather than m3db directly")
 	flag.Parse()
 }
 
 func main() {
+	if coordinator {
+		log.Println("Benchmarking on m3coordinator...")
```
benraskin92: Same as above

benchmark/write/benchmarker.go
arnikola: My bad, changed them to printlns to remove the \n then didn't actually end up removing them

benchmark/common/parse_json.go
arnikola: Would be a separate commit

benchmark/read/read_benchmark.go
arnikola: I think the logger from `/util/logging` has a bunch of additional stuff for id/context tracking, which isn't really needed here

PR #30, opened 2018-03-07
Add search endpoint

benraskin92:


policy/filter/storage_test.go
```diff
@@ -15,6 +15,11 @@ type mockStore struct {
 func (m *mockStore) Fetch(_ context.Context, _ *storage.FetchQuery, _ *storage.FetchOptions) (*storage.FetchResult, error) {
 	return nil, nil
 }
+
+func (m *mockStore) FetchTags(_ context.Context, _ *storage.FetchQuery, _ *storage.FetchOptions) (*storage.SearchResults, error) {
```
nikunjgit: you can probably autogen this

services/m3coordinator/handler/search.go
```diff
@@ -0,0 +1,79 @@
...
+	return &fetchQuery, nil
+}
+
+func (h *SearchHandler) search(reqCtx context.Context, searchReq *storage.FetchQuery, searchOpts *storage.FetchOptions) (*storage.SearchResults, error) {
+	return h.store.FetchTags(reqCtx, searchReq, searchOpts)
+}
+
+func newFetchOptions() *storage.FetchOptions {
+	return &storage.FetchOptions{
+		Limit: 100,
```
nikunjgit: config ?

storage/index.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"github.com/m3db/m3coordinator/models"
+
+	"github.com/m3db/m3db/storage/index"
+	"github.com/m3db/m3ninx/index/segment"
+	"github.com/m3db/m3x/ident"
+)
+
+// FromM3IdentToMetric converts an M3 ident metric to a coordinator metric
```
nikunjgit: does it make sense to always just have it in ident format to avoid this conversion ?

services/m3coordinator/handler/search.go
```diff
@@ -0,0 +1,79 @@
...
+
+// NewSearchHandler returns a new instance of handler
+func NewSearchHandler(storage storage.Storage) http.Handler {
+	return &SearchHandler{store: storage}
+}
+
+func (h *SearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	logger := logging.WithContext(r.Context())
+
+	req, rErr := h.parseRequest(r)
```
arnikola: Can probably just be `err`

services/m3coordinator/handler/search.go
```diff
@@ -0,0 +1,79 @@
...
+
+	var fetchQuery storage.FetchQuery
+	if err := json.Unmarshal(body, &fetchQuery); err != nil {
+		return nil, NewParseError(err, http.StatusBadRequest)
+	}
+
+	return &fetchQuery, nil
+}
+
+func (h *SearchHandler) search(reqCtx context.Context, searchReq *storage.FetchQuery, searchOpts *storage.FetchOptions) (*storage.SearchResults, error) {
```
arnikola: `ctx` `query` `opts`

services/m3coordinator/handler/search.go
arnikola: I think this might work better as a queryarg than a config, so you can dynamically change the limit based on the request

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+		TagMatchers: matchers,
+		Start:       time.Now().Add(-10 * time.Minute),
+		End:         time.Now(),
+	}
+}
+
+func generateSearchBody(t *testing.T) io.Reader {
+	req := generateSearchReq()
+	data, err := json.Marshal(req)
+	if err != nil {
```
arnikola: `require.NoError(t, err)`

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+		t.Fatal("could not marshal json request")
+	}
+
+	return bytes.NewReader(data)
+}
+
+func generateQueryResults(t *testing.T, tagsIter index.TaggedIDsIter) index.QueryResults {
```
arnikola: `t *testing.T` is unnecessary, surprisingly linter was ok with it; also return `*index.QueryResults`

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+	return index.QueryResults{
+		Iter: tagsIter,
+	}
+}
+
+func generateTag() ident.Tag {
```
arnikola: `*ident.Tag`

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+	mockTagIter.EXPECT().Next().Return(true).MaxTimes(1)
+	mockTagIter.EXPECT().Next().Return(false)
+	mockTagIter.EXPECT().Current().Return(generateTag())
+	mockTagIter.EXPECT().Close()
+	mockTagIter.EXPECT().Remaining().Return(0)
+
+	mockTaggedIDsIter := index.NewMockTaggedIDsIter(ctrl)
+	mockTaggedIDsIter.EXPECT().Next().Return(true).MaxTimes(1)
+	mockTaggedIDsIter.EXPECT().Next().Return(false)
+	mockTaggedIDsIter.EXPECT().Current().Return(ident.StringID("test_namespace"), ident.StringID("test_id"), mockTagIter)
```
arnikola: make test_namespace and test_id `const` to avoid reuse when checking

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	storage := local.NewStorage(session, "metrics", resolver.NewStaticResolver(policy.NewStoragePolicy(time.Second, xtime.Second, time.Hour*48)))
+	search := &SearchHandler{store: storage}
+
+	server := httptest.NewServer(search)
+	return server
+}
+
+func TestSearchResponse(t *testing.T) {
+	logging.InitWithCores(nil)
```
arnikola: don't need to do this here, searchServer should handle it

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+	searchServer(t)
+	ctrl := gomock.NewController(t)
+	mockTaggedIDsIter := generateTagIters(ctrl)
+
+	session := client.NewMockSession(ctrl)
+	session.EXPECT().FetchTaggedIDs(gomock.Any(), gomock.Any()).Return(generateQueryResults(t, mockTaggedIDsIter), nil)
+
+	storage := local.NewStorage(session, "metrics", resolver.NewStaticResolver(policy.NewStoragePolicy(time.Second, xtime.Second, time.Hour*48)))
+	searchHandler := &SearchHandler{store: storage}
```
arnikola: You could avoid rewriting most of this if you have `searchServer` return `(*httptest.Server, *SearchHandler)`

services/m3coordinator/handler/search_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func generateQueryResults(t *testing.T, tagsIter index.TaggedIDsIter) index.QueryResults {
+	return index.QueryResults{
+		Iter: tagsIter,
+	}
+}
+
+func generateTag() ident.Tag {
+	return ident.Tag{
+		Name:  ident.StringID("foo"),
```
arnikola: foo and bar could be const

services/m3coordinator/handler/search_test.go
arnikola: Actually, this function doesn't use `server`, maybe refactor `searchServer` to `generateSearchHandler(t *testing.T) *SearchHandler`, then only start the server in `TestSearchEndpoint`?

storage/remote/storage.go
```diff
@@ -2,11 +2,16 @@ package remote
...
 import (
 	"context"
+	"errors"
 
 	"github.com/m3db/m3coordinator/storage"
 	"github.com/m3db/m3coordinator/tsdb/remote"
 )
 
+var (
+	errNotImplemented = errors.New("not implemented")
```
arnikola: Make this a public const and put it into `errors.go`, then your tests can compare with that one rather redefining the error

tsdb/remote/server_test.go
```diff
@@ -83,6 +85,10 @@ func (s *mockStorage) Fetch(ctx context.Context, query *storage.FetchQuery, _ *s
 	}, nil
 }
 
+func (s *mockStorage) FetchTags(ctx context.Context, query *storage.FetchQuery, _ *storage.FetchOptions) (*storage.SearchResults, error) {
+	return nil, errNotImplemented
```
arnikola: Should this one return `nil, nil` and `errStorage` can return the error?

storage/local/storage.go
```diff
@@ -84,6 +85,35 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+	}
+
+	m3query := storage.FetchQueryToM3Query(query)
+	opts := storage.FetchOptionsToM3Options(options, query)
+	results, err := s.session.FetchTaggedIDs(m3query, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	iter := results.Iter
```
arnikola: defer it.Close()

storage/local/storage.go
```diff
@@ -84,6 +85,35 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+
+	var metrics models.Metrics
+	for iter.Next() {
+		metrics = append(metrics, storage.FromM3IdentToMetric(results.Iter.Current()))
```
arnikola: I think this may be a bit clearer as
```
m := storage.FromM3IdentToMetric(iter.Current())
metrics = append(metrics, m)
```

storage/index.go
```diff
@@ -0,0 +1,81 @@
...
+		Filters:     MatchersToFilters(fetchQuery.TagMatchers),
+		Conjunction: segment.AndConjunction, // NB (braskin): & is the only conjunction supported currently
+	}
+	indexQuery.Query = segQuery
+
+	return indexQuery
+}
+
+// MatchersToFilters converts matchers to M3 filters
+func MatchersToFilters(matchers models.Matchers) []segment.Filter {
```
arnikola: []*segment.Filter

storage/index.go
```diff
@@ -0,0 +1,81 @@
...
+	var filters []segment.Filter
+
+	for _, matcher := range matchers {
+		var (
```
arnikola: pull the vars out of the for loop

arnikola: Can do some bit comparisons to make this a bit faster, but dno if this is overengineering it

``` golang
	b1, b2 := byte(1), byte(2)
	for _, matcher := range matchers {
		by := byte(matcher.Type)
		negate = by&b1 == b1
		regexp = by&b2 == b2
		.
		.
		.
	}
```

storage/index.go
```diff
@@ -0,0 +1,81 @@
...
+		Conjunction: segment.AndConjunction, // NB (braskin): & is the only conjunction supported currently
+	}
+	indexQuery.Query = segQuery
+
+	return indexQuery
+}
+
+// MatchersToFilters converts matchers to M3 filters
+func MatchersToFilters(matchers models.Matchers) []segment.Filter {
+	var filters []segment.Filter
```
arnikola: `filters := make([]*segment.Filter, 0, len(matchers))`

storage/index.go
```diff
@@ -0,0 +1,81 @@
...
+	for iterTags.Next() {
+		identTag := iterTags.Current()
+		tags[identTag.Name.String()] = identTag.Value.String()
+	}
+	return tags
+}
+
+// FetchOptionsToM3Options converts a set of coordinator options to M3 options
+func FetchOptionsToM3Options(fetchOptions *FetchOptions, fetchQuery *FetchQuery) index.QueryOptions {
+	return index.QueryOptions{
```
arnikola: *index.QueryOptions

storage/index.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/m3db/m3ninx/index/segment"
+	"github.com/m3db/m3x/ident"
+)
+
+// FromM3IdentToMetric converts an M3 ident metric to a coordinator metric
+func FromM3IdentToMetric(identNamespace, identID ident.ID, iterTags ident.TagIterator) *models.Metric {
+	namespace := identNamespace.String()
+	id := identID.String()
+
+	defer iterTags.Close()
```
arnikola: Close() might be better inside FromIdentToTags

storage/fanout/storage.go
```diff
@@ -59,6 +60,22 @@ func handleFetchResponses(requests []execution.Request) (*storage.FetchResult, e
...
 
+func (s *fanoutStorage) FetchTags(ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (*storage.SearchResults, error) {
+	var metrics models.Metrics
+
+	result := &storage.SearchResults{Metrics: metrics}
+	stores := filterStores(s.stores, s.fetchFilter, query)
+	for _, store := range stores {
+		results, err := store.FetchTags(ctx, query, options)
+		if err != nil {
+			return nil, err
```
arnikola: Better to return `nil, err`, or partial result and err?

storage/index.go
benraskin92: Can't do this because `segment.Query` won't take a pointer

storage/local/storage.go
benraskin92: There's no close for `iter`

services/m3coordinator/handler/search_test.go
benraskin92: I think our convention is to just keep these as is

services/m3coordinator/handler/search_test.go
benraskin92: Per our discussion, can't do this.

services/m3coordinator/handler/search_test.go
benraskin92: Per our discussion, can't return `*index.QueryResults`.

storage/fanout/storage.go
benraskin92: I think `nil, err` cause we've seen issues in the past where users often get confused if we return partial results.

storage/index.go
benraskin92: Can't do this because `FetchTaggedIDs` can't take a pointer.

services/m3coordinator/handler/search.go
```diff
@@ -0,0 +1,97 @@
...
+		Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(jsonData)
+}
+
+func (h *SearchHandler) parseRequest(r *http.Request) (*storage.FetchQuery, *storage.FetchOptions, *ParseError) {
+	defer r.Body.Close()
```
arnikola: Do this after checking that err is not nil, may do something wonky otherwise (like if the body is nil itself or something)

services/m3coordinator/handler/search.go
```diff
@@ -0,0 +1,97 @@
...
+
+// NewSearchHandler returns a new instance of handler
+func NewSearchHandler(storage storage.Storage) http.Handler {
+	return &SearchHandler{store: storage}
+}
+
+func (h *SearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	logger := logging.WithContext(r.Context())
+
+	query, opts, rErr := h.parseRequest(r)
```
arnikola: I think it may be a little cleaner if you parse the body and query strings separately rather than having awkward 3 value return

services/m3coordinator/handler/search.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := json.Unmarshal(body, &fetchQuery); err != nil {
+		return nil, nil, NewParseError(err, http.StatusBadRequest)
+	}
+
+	limitRaw := r.URL.Query().Get("limit")
+	var limit int
+	if limitRaw != "" {
+		limit, err = strconv.Atoi(limitRaw)
+		if err != nil {
+			return nil, nil, NewParseError(err, http.StatusBadRequest)
```
arnikola: I think it's better to just set limit to default if formatting fails

storage/index.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/m3db/m3db/storage/index"
+	"github.com/m3db/m3ninx/index/segment"
+	"github.com/m3db/m3x/ident"
+)
+
+// FromM3IdentToMetric converts an M3 ident metric to a coordinator metric
+func FromM3IdentToMetric(identNamespace, identID ident.ID, iterTags ident.TagIterator) *models.Metric {
+	namespace := identNamespace.String()
+	id := identID.String()
+
```
arnikola: spacing looks a little weird here

storage/index.go
```diff
@@ -0,0 +1,83 @@
...
+// MatchersToFilters converts matchers to M3 filters
+func MatchersToFilters(matchers models.Matchers) []segment.Filter {
+	var (
+		filters []segment.Filter
+		negate  bool
+		regexp  bool
+	)
+
+	for _, matcher := range matchers {
+
```
arnikola: spacing

PR #34, opened 2018-04-04
[WIP] Convert M3DB blocks

benraskin92:


ts/step/iterator.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"errors"
+
+	"github.com/m3db/m3db/encoding"
+)
+
+// SeriesBlockToMultiSeriesBlocks converts M3DB blocks to multi series blocks
+func SeriesBlockToMultiSeriesBlocks(blocks []SeriesBlocks) ([]MultiSeriesBlock, error) {
+	blocksLen := len(blocks[0].Blocks)
```
nikunjgit: maybe call it numBlocks ?

ts/step/iterator.go
```diff
@@ -0,0 +1,40 @@
...
+	if !validateBlockSize(blocks, blocksLen) {
+		return []MultiSeriesBlock{}, errors.New("validation failed. number of blocks is not uniform across SeriesBlocks")
+	}
+
+	multiSeriesBlocks := make([]MultiSeriesBlock, 0, blocksLen)
+	for i := 0; i < blocksLen; i++ {
+		s := make([]encoding.SeriesIterator, 0, len(blocks))
```
nikunjgit: separate out len(blocks) and call it numSeries ?

ts/step/iterator.go
```diff
@@ -0,0 +1,40 @@
...
+		for _, block := range blocks {
+			s = append(s, block.Blocks[i].SeriesIterator)
```
nikunjgit: since you already know the size, instead of append, directly add by index ?

ts/step/iterator.go
```diff
@@ -0,0 +1,40 @@
...
+				Start:           blocks[0].Blocks[i].Start,
+				End:             blocks[0].Blocks[i].End,
+				SeriesIterators: encoding.NewSeriesIterators(s, nil),
+			})
+	}
+
+	return multiSeriesBlocks, nil
+}
+
+func validateBlockSize(blocks []SeriesBlocks, checkingLen int) bool {
```
nikunjgit: you probably also need to ensure blocks are aligned across series ?

ts/step/iterator.go
```diff
@@ -0,0 +1,40 @@
...
+	for i := 0; i < blocksLen; i++ {
+		s := make([]encoding.SeriesIterator, 0, len(blocks))
+		for _, block := range blocks {
+			s = append(s, block.Blocks[i].SeriesIterator)
+		}
+		multiSeriesBlocks = append(multiSeriesBlocks,
+			MultiSeriesBlock{
+				Start:           blocks[0].Blocks[i].Start,
+				End:             blocks[0].Blocks[i].End,
+				SeriesIterators: encoding.NewSeriesIterators(s, nil),
```
nikunjgit: check with @robskillington if we ever need to pass non nil here

ts/step/iterator.go
```diff
@@ -0,0 +1,80 @@
+package step
```
robskillington: Need license header for this file.

ts/step/iterator.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"errors"
+	"time"
+
+	"github.com/m3db/m3db/encoding"
+)
+
+var (
+	errBlocksMisaligned = errors.New("validation failed. blocks are misaligned on either start or end times")
+	errNumBlocks        = errors.New("validation failed. number of blocks is not uniform across SeriesBlocks")
```
robskillington: nit: use commas instead of full stops, each error should be just a sentence if and when possible.

ts/step/iterator.go
robskillington: We should definitely be using a pool if and where possible, can you make this method take a `seriesIteratorsPool MutableSeriesIteratorsPool` and if its non-nil get the series iterators from there.

ts/step/types.go
```diff
@@ -0,0 +1,47 @@
...
+type MultiSeriesBlock struct {
+	Start           time.Time
+	End             time.Time
+	SeriesIterators encoding.SeriesIterators
+}
+
+// Iter iterates through an M3DBStepIterator vertically
+type Iter interface {
+	Next() bool
+	Current() Data
```
robskillington: Hm this interface doesn't match the `ValuesIterator()` interface which gives you a `[]float64`.

ts/step/types.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// M3DBStepIterator implements the Iter interface
+type M3DBStepIterator struct {
```
robskillington: As discussed, lets perhaps use the words `Compressed` vs `M3DB`.

nikunjgit:
We should probably changed it to have the latest design into consideration ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+	errBlocksMisaligned   = errors.New("blocks are misaligned on either start or end times")
+	errNumBlocks          = errors.New("number of blocks is not uniform across SeriesBlocks")
+	errMultipleNamespaces = errors.New("consolidating multiple namespaces is currently not supported")
+)
+
+// SeriesBlockToMultiSeriesBlocks converts M3DB blocks to multi series blocks
+func SeriesBlockToMultiSeriesBlocks(multiNamespaceSeriesList []MultiNamespaceSeries, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) (MultiSeriesBlocks, error) {
+	for _, multiNamespaceSeries := range multiNamespaceSeriesList {
+		numBlocks := len(multiNamespaceSeries[0].Blocks)
+		if err := validateBlocks(multiNamespaceSeries, numBlocks); err != nil {
```
nikunjgit: is this true ? Seems like you're comparing all namespaces for the same series should have the same size blocks

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+			return MultiSeriesBlocks{}, err
+		}
+	}
+
+	var multiSeriesBlocks MultiSeriesBlocks
+	for i, multiNamespaceSeries := range multiNamespaceSeriesList {
```
nikunjgit: instead of i, use a more informative name

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+		multiNSConsolidatedSeriesBlocks, err := newMultiNSConsolidatedSeriesBlocks(multiNamespaceSeries, seriesIteratorsPool)
+		if err != nil {
+			return MultiSeriesBlocks{}, err
+		}
+		if i == 0 {
+			multiSeriesBlocks = make(MultiSeriesBlocks, len(multiNSConsolidatedSeriesBlocks))
+		}
+		for j, multiNSConsolidatedSeriesBlock := range multiNSConsolidatedSeriesBlocks {
```
nikunjgit: instead of j, use a more informative name

ts/iter/types.go
```diff
@@ -0,0 +1,116 @@
...
+	consolidationFunc    ConsolidationFunc
+}
+
+// ConsolidationFunc determines how to consolidate across namespaces
+type ConsolidationFunc func(existing, toAdd float64, count int) float64
+
+// MultiNSConsolidatedSeriesBlocks contain all of the consolidated blocks for
+// a single timeseries across namespaces.
+// Each ConsolidatedBlockIterator will have the same size
+type MultiNSConsolidatedSeriesBlocks []MultiNSConsolidatedSeriesBlock
```
nikunjgit: change this to ConsolidatedSeriesBlocks

ts/iter/types.go
```diff
@@ -0,0 +1,116 @@
...
+// MultiNamespaceSeries is a single timeseries for multiple namespaces
+type MultiNamespaceSeries []SeriesBlocks
+
+// ID enforces the same ID across namespaces
+func (n MultiNamespaceSeries) ID() ident.ID { return n[0].ID }
+
+// ConsolidatedSeriesBlock is a single block for a given timeseries and namespace
+// which contains all of the necessary SeriesIterators so that consolidation can
+// happen across namespaces
+type ConsolidatedSeriesBlock struct {
```
nikunjgit: change this to ConslidatedNSBlock

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	var multiSeriesBlocks MultiSeriesBlocks
+	for i, multiNamespaceSeries := range multiNamespaceSeriesList {
+		multiNSConsolidatedSeriesBlocks, err := newMultiNSConsolidatedSeriesBlocks(multiNamespaceSeries, seriesIteratorsPool)
+		if err != nil {
+			return MultiSeriesBlocks{}, err
+		}
+		if i == 0 {
```
nikunjgit: newline ?

ts/iter/iterator.go
nikunjgit: newline ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+				multiSeriesBlocks[j].Start = multiNSConsolidatedSeriesBlock.Start
+				multiSeriesBlocks[j].End = multiNSConsolidatedSeriesBlock.End
+			}
+			if multiNSConsolidatedSeriesBlock.Start != multiSeriesBlocks[j].Start || multiNSConsolidatedSeriesBlock.End != multiSeriesBlocks[j].End {
+				return MultiSeriesBlocks{}, err
+			}
+			multiSeriesBlocks[j].Blocks = append(multiSeriesBlocks[j].Blocks, multiNSConsolidatedSeriesBlock)
+		}
+	}
+	return multiSeriesBlocks, nil
```
nikunjgit: newline ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func newMultiNSConsolidatedSeriesBlocks(multiNamespaceSeries MultiNamespaceSeries, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) (MultiNSConsolidatedSeriesBlocks, error) {
+	var multiNSConsolidatedSeriesBlocks MultiNSConsolidatedSeriesBlocks
+
+	// todo(braskin): remove this once we support consolidating multiple namespaces
```
nikunjgit: just have //TODO: remove this once we support consolidating multiple namespaces

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+			if multiNSConsolidatedSeriesBlock.Start != multiSeriesBlocks[j].Start || multiNSConsolidatedSeriesBlock.End != multiSeriesBlocks[j].End {
+				return MultiSeriesBlocks{}, err
+			}
+			multiSeriesBlocks[j].Blocks = append(multiSeriesBlocks[j].Blocks, multiNSConsolidatedSeriesBlock)
+		}
+	}
+	return multiSeriesBlocks, nil
+}
+
+func newMultiNSConsolidatedSeriesBlocks(multiNamespaceSeries MultiNamespaceSeries, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) (MultiNSConsolidatedSeriesBlocks, error) {
```
nikunjgit: even though they are private, you would probably need some comments for these functions

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
...
+	var multiNSConsolidatedSeriesBlocks MultiNSConsolidatedSeriesBlocks
+
+	// todo(braskin): remove this once we support consolidating multiple namespaces
+	if len(multiNamespaceSeries) > 1 {
+		return multiNSConsolidatedSeriesBlocks, errMultipleNamespaces
+	}
+
+	for i, seriesBlocks := range multiNamespaceSeries {
+		sliceOfConsolidatedSeriesBlocks := newConsolidatedSeriesBlocks(seriesBlocks, seriesIteratorsPool)
```
nikunjgit: better naming ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,153 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
nikunjgit: might have to call this file something else now since its not really implementing any iterators

ts/iter/iterator.go
benraskin92: this was leftover from the previous version of the diff. I'll remove for now, but we will need to check the block size and alignment before creating the `[]MultiNamespaceSeries`

ts/iter/iterator.go
benraskin92: `blocks.go`?

nikunjgit: maybe it can be ts/m3db/blocks.go ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+	for multiNamespaceSeriesIdx, multiNamespaceSeries := range multiNamespaceSeriesList {
+		consolidatedSeriesBlocks, err := newConsolidatedSeriesBlocks(multiNamespaceSeries, seriesIteratorsPool)
+		if err != nil {
+			return MultiSeriesBlocks{}, err
+		}
+
+		if multiNamespaceSeriesIdx == 0 {
+			multiSeriesBlocks = make(MultiSeriesBlocks, len(consolidatedSeriesBlocks))
+		}
+		for consolidatedSeriesBlockIdx, consolidatedSeriesBlock := range consolidatedSeriesBlocks {
```
nikunjgit: new line after if block ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+			if multiNamespaceSeriesIdx == 0 {
+				multiSeriesBlocks[consolidatedSeriesBlockIdx].Start = consolidatedSeriesBlock.Start
+				multiSeriesBlocks[consolidatedSeriesBlockIdx].End = consolidatedSeriesBlock.End
+			}
+
+			if consolidatedSeriesBlock.Start != multiSeriesBlocks[consolidatedSeriesBlockIdx].Start || consolidatedSeriesBlock.End != multiSeriesBlocks[consolidatedSeriesBlockIdx].End {
+				return MultiSeriesBlocks{}, err
+			}
+			multiSeriesBlocks[consolidatedSeriesBlockIdx].Blocks = append(multiSeriesBlocks[consolidatedSeriesBlockIdx].Blocks, consolidatedSeriesBlock)
```
nikunjgit: new line after if block ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return multiSeriesBlocks, nil
+}
+
+// newConsolidatedSeriesBlocks creates consolidated blocks by timeseries across namespaces
+func newConsolidatedSeriesBlocks(multiNamespaceSeries MultiNamespaceSeries, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) (ConsolidatedSeriesBlocks, error) {
+	var consolidatedSeriesBlocks ConsolidatedSeriesBlocks
+
+	// todo: remove this once we support consolidating multiple namespaces
+	if len(multiNamespaceSeries) > 1 {
```
nikunjgit: we can probably move this to a lower layer

ts/iter/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+		for consolidatedNSBlockIdx, consolidatedNSBlock := range consolidatedNSBlocks {
+			if seriesBlocksIdx == 0 {
+				consolidatedSeriesBlocks[consolidatedNSBlockIdx].Start = consolidatedNSBlock.Start
+				consolidatedSeriesBlocks[consolidatedNSBlockIdx].End = consolidatedNSBlock.End
+			}
+
+			if consolidatedNSBlock.Start != consolidatedSeriesBlocks[consolidatedNSBlockIdx].Start || consolidatedNSBlock.End != consolidatedSeriesBlocks[consolidatedNSBlockIdx].End {
+				return ConsolidatedSeriesBlocks{}, errBlocksMisaligned
+			}
+			consolidatedSeriesBlocks[consolidatedNSBlockIdx].ConsolidatedNSBlocks = append(consolidatedSeriesBlocks[consolidatedNSBlockIdx].ConsolidatedNSBlocks, consolidatedNSBlock)
```
nikunjgit: new line after if block ?

ts/iter/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	return consolidatedSeriesBlocks, nil
+}
+
+// newConsolidatedNSBlocks creates a slice of consolidated blocks per namespace for a single timeseries
+func newConsolidatedNSBlocks(seriesBlocks SeriesBlocks, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) []ConsolidatedNSBlock {
+	var consolidatedNSBlocks []ConsolidatedNSBlock
```
nikunjgit: use make ?

ts/iter/iterator_test.go
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
nikunjgit: probably need to test cases like misaligned blocks, missing blocks,

ts/iter/iterator_test.go
```diff
@@ -0,0 +1,104 @@
...
+const (
+	blockOneID = "test_one"
+	blockTwoID = "test_two"
+)
+
+func newMultiNamespaceSeries(ctrl *gomock.Controller, now time.Time) ([]SeriesBlocks, []SeriesBlocks) {
+	seriesBlocksOne := newM3SeriesBlock(blockOneID, ctrl, now)
+	blocksOne := SeriesBlocks{
+		Blocks: seriesBlocksOne,
+		ID:     ident.StringID(blockOneID),
```
nikunjgit: isn't this the series ID ?

ts/iter/iterator_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/m3db/m3db/encoding"
+	"github.com/m3db/m3x/ident"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	blockOneID = "test_one"
```
nikunjgit: probably have the test pass in the IDs instead of making them a const

ts/iter/types.go
```diff
@@ -0,0 +1,116 @@
...
+	End    time.Time
+	Blocks ConsolidatedSeriesBlocks
+}
+
+// MultiSeriesBlocks is a slice of MultiSeriesBlock
+// todo(braskin): add close method on this to close each SeriesIterator
+type MultiSeriesBlocks []MultiSeriesBlock
+
+// StepIter iterates through a CompressedStepIterator vertically
+type StepIter interface {
```
nikunjgit: i'd say take out the iter stuff from this diff since you're going to rename the package anyway

ts/m3db/blocks.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// SeriesBlockToMultiSeriesBlocks converts M3DB blocks to multi series blocks
+func SeriesBlockToMultiSeriesBlocks(multiNamespaceSeriesList []MultiNamespaceSeries, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) (MultiSeriesBlocks, error) {
+	// todo(braskin): validate blocks size and aligment per namespace before creating []MultiNamespaceSeries
+	var multiSeriesBlocks MultiSeriesBlocks
+	for multiNamespaceSeriesIdx, multiNamespaceSeries := range multiNamespaceSeriesList {
+		consolidatedSeriesBlocks, err := newConsolidatedSeriesBlocks(multiNamespaceSeries, seriesIteratorsPool)
+		if err != nil {
+			return MultiSeriesBlocks{}, err
```
nikunjgit: return a nil slice instead of an empty slice

ts/m3db/types.go
```diff
@@ -0,0 +1,101 @@
...
+// happen across namespaces
+type ConsolidatedNSBlock struct {
+	ID              ident.ID
+	Namespace       string
+	Start           time.Time
+	End             time.Time
+	SeriesIterators encoding.SeriesIterators
+}
+
+func (c ConsolidatedNSBlock) isBeyondBounds(consolidatedSeriesBlock ConsolidatedSeriesBlock) error {
```
nikunjgit: should just return a bool and caller can decide whether to error

ts/m3db/types.go
```diff
@@ -0,0 +1,101 @@
...
+// ConsolidatedSeriesBlock is a single series consolidated across different namespaces
+// for a single block
+type ConsolidatedSeriesBlock struct {
+	Start                time.Time
+	End                  time.Time
+	ConsolidatedNSBlocks []ConsolidatedNSBlock
+	consolidationFunc    ConsolidationFunc
+}
+
+func (c ConsolidatedSeriesBlock) isBeyondBounds(multiSeriesBlock MultiSeriesBlock) error {
```
nikunjgit: should just return a bool and caller can decide whether to error

ts/m3db/types.go
nikunjgit: Also, can you name it to : Contains ?

ts/m3db/types.go
nikunjgit: Also, can you name it to : Contains ?

PR #37, opened 2018-04-10
Add physical plan generator and result node

nikunjgit:


functions/fetch.go
```diff
@@ -21,3 +22,8 @@ type FetchOp struct {
 func (o *FetchOp) OpType() string {
 	return FetchType
 }
+
+// String representation
+func (o *FetchOp) String() string {
```
arnikola: Question rather than comment;

If we implement `String() string`, is it convention to add `Stringer` to the interface?

parser/interface.go
```diff
@@ -16,19 +16,30 @@ type TransformID string
 // Operation is a function that can be applied to data
 type Operation interface {
 	OpType() string
+	String() string
```
arnikola: Stringer

parser/interface.go
```diff
@@ -16,19 +16,30 @@ type TransformID string
...
 }
 
 // Transforms is a slice of Transform
 type Transforms []*Transform
 
-// Transform is a node in common DAG which can be uniquely identified
+// Transform is a node in common DAG which can be uniquely identified. It is immutable
```
arnikola: Transform is an immutable node...

plan/logical.go
```diff
@@ -66,9 +58,47 @@ func GenerateLogicalPlan(transforms parser.Transforms, edges parser.Edges) (*Log
...
+
+func (l *LogicalPlan) String() string {
+	return fmt.Sprintf("Transforms: %s, Pipeline: %s", l.Transforms, l.Pipeline)
+}
+
+// Clone the plan, the transform is immutable so its left as is
+func (l *LogicalPlan) Clone() *LogicalPlan {
+	transforms := make(map[parser.TransformID]*LogicalStep)
+	for id, step := range l.Transforms {
+		transforms[id] = step.Clone()
```
arnikola: Do you need to clone if `step` is immutable?

plan/logical.go
```diff
@@ -19,39 +19,31 @@ type LogicalStep struct {
 	Transform *parser.Transform
 }
 
-// NewLogicalPlan returns an empty logical plan
-func NewLogicalPlan() *LogicalPlan {
+// newLogicalPlan returns an empty logical plan
+func newLogicalPlan() *LogicalPlan {
 	return &LogicalPlan{
 		Transforms: make(map[parser.TransformID]*LogicalStep),
```
arnikola: Rename to Steps

plan/physical.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+
+	"github.com/m3db/m3coordinator/parser"
+	"github.com/m3db/m3coordinator/storage"
+)
+
+// Physical represents the physical plan
+type Physical struct {
+	Transforms map[parser.TransformID]*LogicalStep
```
arnikola: Steps

functions/fetch.go
nikunjgit: yeah good point

plan/physical.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"fmt"
+
+	"github.com/m3db/m3coordinator/parser"
+	"github.com/m3db/m3coordinator/storage"
+)
+
+// Physical represents the physical plan
+type Physical struct {
```
arnikola: PhysicalPlan?

plan/physical.go
```diff
@@ -0,0 +1,72 @@
...
+	cloned := lp.Clone()
+	return &Physical{
+		Transforms: cloned.Transforms,
+		Pipeline:   cloned.Pipeline,
+	}
+}
+
+// PhysicalPlan is used to generate a physical plan. Its responsibilities include creating consolidation nodes, result nodes,
+// pushing down predicates, changing the ordering for nodes
+func PhysicalPlan(lp *LogicalPlan, storage storage.Storage) (*Physical, error) {
```
arnikola: NewPhysicalPlan

plan/physical.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Physical represents the physical plan
+type Physical struct {
+	Transforms map[parser.TransformID]*LogicalStep
+	Pipeline   []parser.TransformID // Ordered list of steps to be performed
+	ResultStep *LogicalStep
+}
+
+// newPhysicalPlan generates a new physical plan after cloning the logical plan so that any changes here do not update the logical plan
+func newPhysicalPlan(lp *LogicalPlan) *Physical {
```
arnikola: You're only doing this once, might as well inline?

plan/physical.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (p *Physical) createResultNode() error {
+	leaf, err := p.leafNode()
+	if err != nil {
+		return err
+	}
+
+	resultNode := parser.NewTransformFromOperation(&ResultOp{}, len(p.Transforms)+1)
+	resultStep := newLogicalStep(resultNode)
+	resultStep.Parents = append(resultStep.Parents, leaf.Transform.ID())
```
arnikola: add a transformID() method to leafNode that would expose underlying transformID?

plan/physical.go
nikunjgit: umm putting clone inside the function which needs a new physical plan might be weird.

plan/physical.go
```diff
@@ -0,0 +1,72 @@
...
+	p.ResultStep = resultStep
+	return nil
+}
+
+func (p *Physical) leafNode() (*LogicalStep, error) {
```
arnikola: Can we save on this by having the logical plan store any transforms it adds without children?

plan/physical_test.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	lp, err := GenerateLogicalPlan(transforms, edges)
+	require.NoError(t, err)
+	p, err := PhysicalPlan(lp, nil)
+	require.NoError(t, err)
+	node, err := p.leafNode()
+	require.NoError(t, err)
+	assert.Equal(t, node.Transform.ID(), countTransform.ID())
+	assert.Equal(t, p.ResultStep.Transform.Op().OpType(), ResultType)
```
arnikola: Possibly a TransformOpType() function on Transform too?

plan/physical.go
nikunjgit: that might be non trivial since logical plan creates children. So I suppose it would still need to loop through the list once all children are created

plan/physical_test.go
nikunjgit: haven't being using it much, will add it if that becomes the case

plan/logical.go
```diff
@@ -19,56 +19,91 @@ type LogicalStep struct {
 	Transform *parser.Transform
 }
 
-// NewLogicalPlan returns an empty logical plan
-func NewLogicalPlan() *LogicalPlan {
+// newLogicalPlan returns an empty logical plan
+func newLogicalPlan() *LogicalPlan {
```
arnikola: do we need these private constructors that don't really do anything?

nikunjgit: i can take it out

plan/logical.go
```diff
@@ -42,52 +42,57 @@ func GenerateLogicalPlan(transforms parser.Transforms, edges parser.Edges) (*Log
...
+		Pipeline: pipeline,
 	}
 }
 
 func (l *LogicalStep) String() string {
 	return fmt.Sprintf("Parents: %s, Children: %s, Transform: %s", l.Parents, l.Children, l.Transform)
 }
 
+// ID is a convenience method to expose the inner transforms' ID
+func (l *LogicalStep) ID() parser.TransformID {
```
arnikola: nit: Maybe TransformID()?

plan/logical.go
```diff
@@ -8,8 +8,8 @@ import (
 
 // LogicalPlan converts a DAG into a list of steps to be executed in order
 type LogicalPlan struct {
-	Transforms map[parser.TransformID]*LogicalStep
-	Pipeline   []parser.TransformID // Ordered list of steps to be performed
+	Steps    map[parser.TransformID]*LogicalStep
```
arnikola: I think it would be better to hide implementation details fo Steps behind a basic struct? Exposing map like this seems a little weird

plan/logical.go
```diff
@@ -19,56 +19,81 @@ type LogicalStep struct {
...
-func (l *LogicalPlan) String() string {
-	return fmt.Sprintf("Transforms: %s, Pipeline: %s", l.Transforms, l.Pipeline)
-}
-
-func (l *LogicalStep) String() string {
-	return fmt.Sprintf("Parents: %s, Children: %s, Transform: %s", l.Parents, l.Children, l.Transform)
-}
-
 // GenerateLogicalPlan creates a plan from the DAG structure
 func GenerateLogicalPlan(transforms parser.Transforms, edges parser.Edges) (*LogicalPlan, error) {
```
arnikola: NewLogicalPlan?

plan/logical.go
nikunjgit: the step inherits the ID from transform so I just kept it ID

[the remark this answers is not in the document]
nikunjgit: the step inherits the ID from transform so I just kept it ID(). Outside functions don't really need to know where the ID is coming from

plan/logical.go
nikunjgit: we don't really expect functions to be defined on Steps apart from simple lookups. I thought you didn't like structs which just did one thing :P In the future, I can look at whether Steps even needs to be exposed outside. But too early now

parser/promql/parse.go
```diff
@@ -51,8 +51,8 @@ func walk(node pql.Node) (parser.Transforms, parser.Edges, error) {
 
 		opTransform := parser.NewTransformFromOperation(NewOperator(n.Op), len(transforms))
 		edges = append(edges, &parser.Edge{
-			ParentID: transforms[len(transforms)-1].ID,
-			ChildID:  opTransform.ID,
+			ParentID: transforms[len(transforms)-1].ID(),
```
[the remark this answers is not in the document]
nikunjgit: yeah I suppose I can hide ParentId and ChildId and have accessors but don't struct is super simple so might be an overkill ?

parser/interface.go
```diff
@@ -15,20 +15,31 @@ type TransformID string
...
 type Operation interface {
+	fmt.Stringer
 	OpType() string
 }
 
 // Transforms is a slice of Transform
 type Transforms []*Transform
 
-// Transform is a node in common DAG which can be uniquely identified
+// Transform is an immutable node in common DAG which can be uniquely identified. In the future we would want it to be serializable
```
arnikola: // Transform represents an immutable node in the common DAG with a unique identifier.
// TODO: make this serializable

parser/interface.go
```diff
@@ -15,20 +15,31 @@ type TransformID string
...
 type Transform struct {
-	ID TransformID
-	Op Operation
+	id TransformID
+	op Operation
+}
+
+// ID is a unique ID for the transform
```
arnikola: the unique ID

parser/interface.go
```diff
@@ -15,20 +15,31 @@ type TransformID string
...
+func (t *Transform) ID() TransformID {
+	return t.id
+}
+
+// Op is a operation for the transform
```
arnikola: Op represents the operation used in the transform

parser/promql/parse.go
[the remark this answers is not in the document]
arnikola: Accessors is fine

plan/logical.go
arnikola: Hide behind accessors then?

plan/logical.go
```diff
@@ -19,56 +19,81 @@ type LogicalStep struct {
...
+		}
+		lp.Pipeline = append(lp.Pipeline, transform.ID())
 	}
 
 	// Link all parent/children
 	for _, edge := range edges {
-		parent, ok := lp.Transforms[edge.ParentID]
+		parent, ok := lp.Steps[edge.ParentID]
 		if !ok {
 			return nil, fmt.Errorf("invalid DAG found, parent %s not found", edge.ParentID)
```
arnikola: Maybe include child that needs that parent? Same for other error

plan/logical.go
```diff
@@ -19,56 +19,81 @@ type LogicalStep struct {
...
+
+func (l *LogicalPlan) String() string {
+	return fmt.Sprintf("Steps: %s, Pipeline: %s", l.Steps, l.Pipeline)
+}
+
+// Clone the plan
+func (l *LogicalPlan) Clone() *LogicalPlan {
+	steps := make(map[parser.TransformID]*LogicalStep)
+	for id, step := range l.Steps {
+		steps[id] = step.Clone()
```
arnikola: Aren't all of these immutable at this point?

Parents   []parser.TransformID
Children  []parser.TransformID		 	
Transform *parser.Transform

plan/logical.go
nikunjgit: yeah, added

plan/logical.go
nikunjgit: well they are mutable but the copy of logical plan which physical plan gets will not be able to change the actual logical plan. We currently had bugs in existing query service where we lose the actual query/context so I made a clone to be safe since physical plan will update some parent/child relations.

PR #39, opened 2018-04-17
Namespace endpoints; restructure boot up for KV endpoints

justinjc:
- New namespace get/add/delete endpoints
- Add tests
- Use more updated m3db/m3db version to pick up embedded KV changes
- **Restructure m3coordinator boot up to enable KV endpoints before M3DB topology is set**


services/m3coordinator/handler/placement_init_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	mockClient.EXPECT().Services(gomock.Not(nil)).Return(mockServices, nil).AnyTimes()
+	mockServices.EXPECT().PlacementService(gomock.Not(nil), gomock.Not(nil)).Return(mockPlacementService, nil).AnyTimes()
+
+	handler := NewPlacementInitHandler(mockClient)
+
+	// Test placement init success
+	w := httptest.NewRecorder()
+
+	// Actual JSON passed in does not matter because we're mocking the response out from the PlacementService
```
arnikola: Would it be worth telling the mock to expect the processed JSON sent in?

services/m3coordinator/handler/admin_common.go
```diff
@@ -1,11 +1,18 @@
...
+	"fmt"
+	"time"
+
 	m3clusterClient "github.com/m3db/m3cluster/client"
 	"github.com/m3db/m3cluster/generated/proto/placementpb"
+	"github.com/m3db/m3cluster/kv"
 	"github.com/m3db/m3cluster/placement"
 	"github.com/m3db/m3cluster/services"
 	"github.com/m3db/m3cluster/shard"
+	"github.com/m3db/m3coordinator/services/m3coordinator/config"
```
nikunjgit: import order ?

services/m3coordinator/handler/namespace_delete.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, md := range currentMetadata {
+		if md.ID().String() == r.Name {
+			found = true
+			continue
+		}
+		newMds = append(newMds, md)
+	}
+
+	if !found {
+		return errors.New("unable to find a namespace with specified name")
```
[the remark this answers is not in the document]
justinjc: I think so - do you think returning `200 OK` is more appropriate in this case? I've factored this "not found" error out to return back `400 Bad Request` instead of `500 Internal Server Error` to make it a little clearer.

services/m3coordinator/handler/admin_common.go
```diff
@@ -17,33 +24,58 @@ const (
...
 	}
 
+	var (
+		serviceName        string
+		serviceEnvironment string
+		serviceZone        string
+	)
+
+	if cfg.M3DBClientCfg.EnvironmentConfig.Service == nil {
+		serviceName = DefaultServiceName
```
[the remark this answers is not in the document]
justinjc: If these variables are present in the config, then I'll use them. However, I think it will be common enough for smaller consumers (or people just testing it out) to only have one environment and zone. As such, it seems unnecessary for them to have to specify these values.

generated/proto/admin/namespace.proto
```diff
@@ -0,0 +1,31 @@
...
+
+message NamespaceAddRequest {
+  // Required fields
+  string name = 1;
+  string retention_period = 2;
+  string block_size = 3;
+  string buffer_future = 4;
+  string buffer_past = 5;
+
+  // Optional fields that are defaulted
```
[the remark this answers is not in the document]
justinjc: I don't think the ordering of the fields matter here, right? I've grouped them this way just so that it's clearer for developers. In the future, something like this should be fine:
```
// Required fields
...
string buffer_past = 5;
bool new_required_var = 13;

// Optional fields that are defaulted
bool block_data_expiry = 6;
...
```

services/m3coordinator/handler/namespace_get_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	resp := w.Result()
+	body, _ := ioutil.ReadAll(resp.Body)
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.Equal(t, "{\"registry\":{}}", string(body))
+
+	// Test namespace present
+	w = httptest.NewRecorder()
+
+	req = httptest.NewRequest("GET", "/namespace/get", nil)
```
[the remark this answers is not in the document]
justinjc: At this layer, we've already gone past the routing logic, so actually the endpoint used here does not really matter. Testing for that should be covered at integration testing level - something I will get to later.

services/m3coordinator/handler/placement_delete_test.go
```diff
@@ -0,0 +1,56 @@
...
+	require.NotNil(t, req)
+	mockPlacementService.EXPECT().Delete()
+	handler.ServeHTTP(w, req)
+
+	resp := w.Result()
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	// Test delete error
+	w = httptest.NewRecorder()
+	req = httptest.NewRequest("POST", "/placement/delete", nil)
```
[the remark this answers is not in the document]
justinjc: Similar to above -
> At this layer, we've already gone past the routing logic, so actually the endpoint used here does not really matter. Testing for that should be covered at integration testing level - something I will get to later.

services/m3coordinator/handler/namespace_get_test.go
```diff
@@ -0,0 +1,84 @@
...
+	require.NotNil(t, req)
+
+	registry := nsproto.Registry{
+		Namespaces: map[string]*nsproto.NamespaceOptions{
+			"test": &nsproto.NamespaceOptions{
+				NeedsBootstrap:      true,
+				NeedsFlush:          true,
+				WritesToCommitLog:   true,
+				NeedsFilesetCleanup: false,
```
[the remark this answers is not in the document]
justinjc: As mentioned above, all fields are now required.

services/m3coordinator/handler/namespace_add.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return *protoRegistry, nil
+}
+
+func metadataFromRequest(r *admin.NamespaceAddRequest) (namespace.Metadata, error) {
+	// Explicitly check existence of name. Other required fields are `time.Duration`s,
+	// which will fail to parse as such on empty string.
+	if r.Name == "" {
+		return nil, errors.New("must specify namespace name")
```
benraskin92: Add this error as a var (same for other errors) at top of file

storage/index.go
```diff
@@ -22,12 +22,9 @@ func FromM3IdentToMetric(identNamespace, identID ident.ID, iterTags ident.TagIte
...
-func FromIdentTagsToTags(iterTags ident.TagIterator) models.Tags {
-	defer iterTags.Close()
-
-	tags := make(models.Tags, iterTags.Remaining())
-	for iterTags.Next() {
-		identTag := iterTags.Current()
+func FromIdentTagsToTags(identTags ident.Tags) models.Tags {
+	tags := make(models.Tags, len(identTags))
+	for _, identTag := range identTags {
 		tags[identTag.Name.String()] = identTag.Value.String()
```
arnikola: You may need to Finalize() the identTag here

services/m3coordinator/handler/common.go
```diff
@@ -71,9 +73,9 @@ func ParseRequestParams(r *http.Request) (*RequestParams, error) {
 	return &params, nil
 }
 
-// WriteJSONResponse writes a protobuf message to the ResponseWriter
-func WriteJSONResponse(w http.ResponseWriter, resp interface{}, logger *zap.Logger) {
-	jsonData, err := json.Marshal(resp)
+// WriteJSONResponse writes a generic to the ResponseWriter
```
arnikola: nit `writes generic data`

justinjc:
ci is failing seemingly because https://golang.org/x/lint/golint is currently returning a `404`. Issue here:
https://github.com/golang/lint/issues/397

services/m3coordinator/handler/admin_common.go
```diff
@@ -21,11 +21,18 @@
 package handler
```
nikunjgit: move this to admin package ? You can probably that in the other PR

services/m3coordinator/handler/admin_common.go
```diff
@@ -37,26 +44,40 @@ const (
...
 // PlacementService gets a placement service from an m3cluster client
-func PlacementService(clusterClient m3clusterClient.Client) (placement.Service, error) {
-	cs, err := clusterClient.Services(services.NewOptions())
+func PlacementService(clusterClient m3clusterClient.Client, cfg config.Configuration) (placement.Service, error) {
+	cs, err := clusterClient.Services(services.NewOverrideOptions())
 	if err != nil {
 		return nil, err
 	}
 
+	serviceName := DefaultServiceName
```
nikunjgit: move this outside?  You can probably that in the other PR

justinjc:
Added copyright headers. Will restructure directories for handlers in PR #41.

services/m3coordinator/handler/namespace_get_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/m3db/m3cluster/client"
+	"github.com/m3db/m3cluster/kv"
+	nsproto "github.com/m3db/m3db/generated/proto/namespace"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func SetupNamespaceTest(t *testing.T) (*client.MockClient, *kv.MockStore, *gomock.Controller) {
```
arnikola: nit: Can this be private?

services/m3coordinator/handler/placement_get_test.go
```diff
@@ -35,7 +39,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestPlacementGetHandler(t *testing.T) {
+func SetupPlacementTest(t *testing.T) (*client.MockClient, *placement.MockService) {
```
arnikola: I think this can be private too

services/m3coordinator/handler/placement_get_test.go
```diff
@@ -45,19 +51,65 @@ func TestPlacementGetHandler(t *testing.T) {
...
+	mockClient.EXPECT().Services(gomock.Not(nil)).Return(mockServices, nil).AnyTimes()
+
+	return mockClient, mockPlacementService
+}
+
+func TestPlacementGetHandler(t *testing.T) {
+	mockClient, mockPlacementService := SetupPlacementTest(t)
 	handler := NewPlacementGetHandler(mockClient)
+
+	// Test successful get
```
arnikola: Maybe split into 2 different tests?

services/m3coordinator/httpd/handler.go
```diff
@@ -66,16 +76,28 @@ func NewHandler(storage storage.Storage, engine *executor.Engine, clusterClient
...
 
 // RegisterRoutes registers all http routes.
 func (h *Handler) RegisterRoutes() {
 	logged := withResponseTimeLogging
-	h.Router.HandleFunc(handler.PromReadURL, logged(handler.NewPromReadHandler(h.engine)).ServeHTTP).Methods("POST")
-	h.Router.HandleFunc(handler.PromWriteURL, logged(handler.NewPromWriteHandler(h.storage)).ServeHTTP).Methods("POST")
-	h.Router.HandleFunc(handler.SearchURL, logged(handler.NewSearchHandler(h.storage)).ServeHTTP).Methods("POST")
+
+	promReadHandler := handler.NewPromReadHandler(h.engine)
+	h.lazyHandlers.promRead = promReadHandler.(*handler.PromReadHandler)
```
arnikola: small nit: take it or leave it

In `read.go` can we do something like 

```golang
// PromReadHandler represents a handler for prometheus read endpoint.
type PromReadHandler struct {
	http.Handler
	engine *executor.Engine
}

// NewPromReadHandler returns a new instance of handler.
func NewPromReadHandler(engine *executor.Engine) *PromReadHandler {
	return &PromReadHandler{engine: engine}
}
```
to avoid having to do a cast here? I think it should work the same. Can take a similar approach for the other handlers, tpp

PR #40, opened 2018-04-19
Execution state for query nodes

nikunjgit:


executor/state.go
```diff
@@ -0,0 +1,93 @@
...
+		plan: plan,
+	}
+
+	resultNode, _, err := state.createNode(result)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(state.sources) == 0 {
+		return nil, fmt.Errorf("empty sources for the execution state")
```
benraskin92: Can this just be `errors.New("empty sources for the execution state")`?

executor/state.go
```diff
@@ -0,0 +1,93 @@
...
+	state.resultNode = resultNode
+	return state, nil
+}
+
+// createNode helps to create an execution node recursively
+// TODO: consider modifying this function so that ExecutionState can have a non pointer receiver
+func (s *ExecutionState) createNode(step plan.LogicalStep) (parser.OpNode, *parser.TransformController, error) {
+	stepNode, controller := step.Transform.Node()
+	// TODO: consider using a registry instead of casting to an interface
+	source, ok := stepNode.(Source)
```
benraskin92: error out if `!ok`?

executor/state_test.go
```diff
@@ -0,0 +1,61 @@
...
+	require.NoError(t, err)
+	p, err := plan.NewPhysicalPlan(lp, nil)
+	require.NoError(t, err)
+	_, err = GenerateExecutionState(p)
+	assert.Error(t, err)
+}
+
+
+func TestOnlySources(t *testing.T) {
+	fetchTransform := parser.NewTransformFromOperation(functions.FetchOp{}, 1)
```
benraskin92: Should you test having multiple sources?

executor/state.go
```diff
@@ -0,0 +1,131 @@
...
+	for idx, source := range s.sources {
+		requests[idx] = sourceRequest{source}
+	}
+
+	return execution.ExecuteParallel(ctx, requests)
+}
+
+// String representation of the state
+func (s *ExecutionState) String() string {
+	return fmt.Sprintf("plan : %s\nsources: %s\nresult: %s", s.plan, s.sources, s.resultNode)
```
benraskin92: nit: remove extra space after `plan`

executor/state_test.go
```diff
@@ -0,0 +1,86 @@
...
+	countTransform := parser.NewTransformFromOperation(functions.CountOp{}, 2)
+	transforms := parser.Nodes{fetchTransform, countTransform}
+	edges := parser.Edges{
+		parser.Edge{
+			ParentID: fetchTransform.ID,
+			ChildID:  countTransform.ID,
+		},
+	}
+
+	lp, err := plan.NewLogicalPlan(transforms, edges)
```
benraskin92: Do we want to test the logical and physical plan code here?

executor/transform/controller.go
```diff
@@ -0,0 +1,14 @@
+package transform
+
+import "github.com/m3db/m3coordinator/parser"
```
benraskin92: nit: 
```
import (
	"github.com/m3db/m3coordinator/parser"
)
```

plan/logical_test.go
```diff
@@ -33,3 +33,52 @@ func TestSingleChildParentRelation(t *testing.T) {
...
+	assert.Len(t, lp.Steps[fetchTransform.ID].Children, 2)
+	assert.Len(t, lp.Steps[fetchTransform.ID].Parents, 0)
+	assert.Len(t, lp.Steps[countTransform2.ID].Parents, 1)
+	assert.Equal(t, lp.Steps[countTransform1.ID].Parents[0], lp.Steps[countTransform2.ID].Parents[0])
+}
+
+func TestMultiParent(t *testing.T) {
+	fetchTransform1 := parser.NewTransformFromOperation(functions.FetchOp{}, 1)
+	fetchTransform2 := parser.NewTransformFromOperation(functions.FetchOp{}, 2)
+	countTransform := parser.NewTransformFromOperation(functions.CountOp{}, 3)
```
benraskin92: Can you make a note to change `Count` to something like `asPercent` in the future? would make this example a little clearer.

executor/state_test.go
```diff
@@ -0,0 +1,86 @@
...
+	edges := parser.Edges{}
+	lp, err := plan.NewLogicalPlan(transforms, edges)
+	require.NoError(t, err)
+	p, err := plan.NewPhysicalPlan(lp, nil)
+	require.NoError(t, err)
+	_, err = GenerateExecutionState(p, nil)
+	assert.Error(t, err)
+}
+
+
```
benraskin92: nit: remove newline

executor/state.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+
+// Source represents data sources which are handled differently than other transforms as they are always independent and can always be parallelized
+type Source interface {
+	Execute(ctx context.Context) error
+}
+
+// ExecutionState represents the execution hierarchy
+type ExecutionState struct {
```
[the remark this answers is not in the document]
nikunjgit: that can be ExecutionStatus ?

executor/state_test.go
nikunjgit: ideally we should be able to get the execution state without calling physical and logical plan code. Once this is more stable, we can add dummy physical and logical plans which can be used here directly.

executor/result.go
```diff
@@ -0,0 +1,6 @@
+package executor
```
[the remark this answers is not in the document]
nikunjgit: let me create another diff and add licenses to all files. Then we can start enforcing

plan/physical.go
```diff
@@ -71,6 +69,13 @@ func (p PhysicalPlan) leafNode() (LogicalStep, error) {
 	return leaf, nil
 }
 
+// Step gets the logical step using its unique ID in the DAG
+func (p PhysicalPlan) Step(ID parser.NodeID) (LogicalStep, bool) {
+	// Editor complains when inlining the map get
+	step, ok := p.steps[ID]
```
benraskin92: Meaning you can't just do `return p.steps[ID]`?

nikunjgit: yea, maybe the compiler would be fine but editor complains

PR #41, opened 2018-04-20
Restructure handlers; new placement endpoints; async m3db sessions

justinjc:
Changes:
1. New endpoints for `placement/add` and `placement/remove`
1. Restructured handlers directory and packages into separate `placement` and `namespace` packages, and renamed files/variables accordingly
1. A new wrapper around an m3db session which allows it to load asynchronously without blocking on topology. This enables KV commands to function before topology is set.

---

Adding two more endpoints to add and remove instances from a placement. I've largely recreated these endpoints based on `m3admin`, for which I have two questions:

1. Providing multiple instances to `placement/add` will only add _one_ instance from the provided list. Should it try to add all provided instances instead?

1. `placement/remove` sounds very similar to `placement/delete`, but one is for removing the specified instances from the placement, and the other is to delete the whole placement (including shard the replication factor info). Should we rename these to something clearer?


services/m3coordinator/handler/namespace_delete.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return err
+	}
+
+	currentMetadata, err := currentNamespaceMetadata(kv)
+	if err != nil {
+		return err
+	}
+
+	newMds := []namespace.Metadata{}
```
nikunjgit: can you do a copy, and then find the index and remove using :  
a[i] = a[len(a)-1] 
a = a[:len(a)-1]

services/m3coordinator/handler/namespace_delete.go
```diff
@@ -0,0 +1,123 @@
...
+			continue
+		}
+		newMds = append(newMds, md)
+	}
+
+	if !found {
+		return errNamespaceNotFound
+	}
+
+	// If metadatas are empty, remove the key
```
nikunjgit: how will we make this apparent to the caller of the endpoint

services/m3coordinator/handler/namespace_get.go
```diff
@@ -0,0 +1,74 @@
...
+	} else if err != nil {
+		return emptyReg, err
+	}
+
+	var protoRegistry nsproto.Registry
+
+	if err := value.Unmarshal(&protoRegistry); err != nil {
+		return emptyReg, fmt.Errorf("failed to parse namespace version %v: %v", value.Version(), err)
+	}
+	return protoRegistry, nil
```
nikunjgit: new line ?

services/m3coordinator/handler/placement_get.go
```diff
@@ -54,19 +56,14 @@ func (h *PlacementGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
 }
 
-func (h *PlacementGetHandler) placementGet(ctx context.Context) (placement.Placement, int, error) {
-	ps, err := PlacementService(h.clusterClient)
-	if err != nil {
-		return nil, 0, err
-	}
-
-	placement, version, err := ps.Placement()
+func (h *placementGetHandler) placementGet(ctx context.Context) (placement.Placement, int, error) {
```
nikunjgit: where are you using the ctx ?

services/m3coordinator/handler/namespace_get.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	resp := &admin.NamespaceGetResponse{
+		Registry: &nsRegistry,
+	}
+
+	WriteProtoMsgJSONResponse(w, resp, logger)
+}
+
+func (h *namespaceGetHandler) namespaceGet(ctx context.Context) (nsproto.Registry, error) {
```
nikunjgit: remove ctx ?

services/m3coordinator/handler/namespace_get.go
```diff
@@ -0,0 +1,74 @@
...
+	var emptyReg = nsproto.Registry{}
+	store, err := h.clusterClient.KV()
+	if err != nil {
+		return emptyReg, err
+	}
+
+	value, err := store.Get(M3DBNodeNamespacesKey)
```
nikunjgit: does it make sense to have a helper for 404 case ? You seem to be rewriting this logic at a lot of places

services/m3coordinator/handler/namespace_get.go
```diff
@@ -0,0 +1,74 @@
...
+	if err == kv.ErrNotFound {
+		// Having no namespace should not be treated as an error
+		return emptyReg, nil
+	} else if err != nil {
+		return emptyReg, err
+	}
+
+	var protoRegistry nsproto.Registry
+
```
nikunjgit: Should there be a new line here ?

services/m3coordinator/handler/placement_add.go
```diff
@@ -0,0 +1,96 @@
+package handler
```
nikunjgit: put this in handler/namespace/add.go ?

services/m3coordinator/handler/namespace_delete.go
justinjc: The caller shouldn't care whether we remove the key or not - the effect of removing their specified namespace should be the same.

asyncsession/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+// Calls to methods while uninitialized will return an uninitialized error. The done channel
+// is to notify the caller that the session has finished _attempting_ to get initialized.
+type Session struct {
+	session client.Session
+	done    chan struct{}
+	err     error
+}
+
+// NewSession returns a new Session
+func NewSession(c client.Client, done chan struct{}) *Session {
```
arnikola: Is `done` ever not nil?

asyncsession/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+		asyncSession.session = session
+		asyncSession.err = nil
+	}()
+
+	return asyncSession
+}
+
+// Write value to the database for an ID
+func (s *Session) Write(namespace, id ident.ID, t time.Time, value float64, unit xtime.Unit, annotation []byte) error {
+	if s.err != nil {
```
arnikola: Should these make an attempt to wait for session load?

asyncsession/asyncsession.go
justinjc: I initially did not plan on having this channel, but I use it to guarantee testing scenarios, e.g. [here](https://github.com/m3db/m3coordinator/pull/41/files#diff-e1297f9813f6ee7bf4fd4eb0c18a9d20R61). I understand that using the channel kind of defeats the purpose of this construct though.

services/m3coordinator/handler/namespace/add.go
```diff
@@ -42,74 +42,74 @@ import (
...
 	}
 	defer r.Body.Close()
 
 	addReq := new(admin.NamespaceAddRequest)
 	if err := json.Unmarshal(body, addReq); err != nil {
-		return nil, NewParseError(err, http.StatusBadRequest)
+		return nil, handler.NewParseError(err, http.StatusBadRequest)
+	}
+
+	if addReq.Name == "" || addReq.RetentionPeriod == "" || addReq.BlockSize == "" || addReq.BufferFuture == "" ||
```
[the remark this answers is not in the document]
justinjc: Sure. This sounds generic enough so I put this in the `functions` package. Named it `HasEmptyString` instead to be clear.

services/m3coordinator/handler/namespace/common.go
```diff
@@ -0,0 +1,63 @@
...
+type Handler struct {
+	store kv.Store
+}
+
+// Metadata returns the current metadata in the given store
+func Metadata(store kv.Store) ([]namespace.Metadata, error) {
+	value, err := store.Get(M3DBNodeNamespacesKey)
+	if err != nil {
+		if err == kv.ErrNotFound {
+			return []namespace.Metadata{}, nil
```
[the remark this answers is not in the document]
justinjc: I think in the general case it would be better for the caller to handle this, but being in the `namespace` package within handlers, it makes sense semantically that having no Metadata set in KV is equal just empty Metadata instead of an error.

services/m3coordinator/handler/namespace/common_test.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+	}
+	mockMetaValue := kv.NewMockValue(ctrl)
+	mockMetaValue.EXPECT().Unmarshal(gomock.Not(nil)).Return(nil).SetArg(0, registry)
+
+	// Test namespaces
+	mockKV.EXPECT().Get(M3DBNodeNamespacesKey).Return(mockMetaValue, nil)
+	meta, err = Metadata(mockKV)
+	assert.NotNil(t, meta)
+	assert.Equal(t, 2, len(meta))
```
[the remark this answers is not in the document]
justinjc: Interesting, didn't know this one! It's odd that the "expected" value is the third argument instead of the second, which is the traditional place...

services/m3coordinator/handler/namespace/delete.go
```diff
@@ -18,123 +18,118 @@
...
 
-func (h *namespaceDeleteHandler) namespaceDelete(ctx context.Context, r *admin.NamespaceDeleteRequest) error {
-	kv, err := h.clusterClient.KV()
-	if err != nil {
-		return err
-	}
-
-	currentMetadata, err := currentNamespaceMetadata(kv)
+func (h *deleteHandler) delete(r *admin.NamespaceDeleteRequest) error {
+	metadatas, err := Metadata(h.store)
```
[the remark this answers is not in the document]
justinjc: `Metadata(...)` essentially gets a value from KV and translates it to an M3DB Metadata. The metadata can be different by the time this function gets called again.

services/m3coordinator/handler/namespace/delete.go
```diff
@@ -18,123 +18,118 @@
...
+	// If we are going to delete the last remaining metadata, remove the key
+	if len(metadatas) == 1 {
+		if _, err = h.store.Delete(M3DBNodeNamespacesKey); err != nil {
 			return fmt.Errorf("unable to delete kv key: %v", err)
 		}
 
 		return nil
 	}
 
+	// Replace the index where we found the metadata with the last element, then truncate
```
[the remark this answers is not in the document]
justinjc: The ordering of namespaces shouldn't matter. I think when @nikunjgit suggested this, he was thinking of the performance improvement of swapping and truncating, as you wouldn't have to recopy the rest of the slice.

services/m3coordinator/handler/namespace/delete.go
```diff
@@ -18,123 +18,118 @@
...
+func (h *deleteHandler) delete(r *admin.NamespaceDeleteRequest) error {
+	metadatas, err := Metadata(h.store)
 	if err != nil {
 		return err
 	}
 
-	newMds := []namespace.Metadata{}
-	found := false
-	for _, md := range currentMetadata {
+	mdIdx := -1
```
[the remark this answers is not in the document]
justinjc: Namespace names are unique and stopping after finding the first matching name should suffice.

asyncsession/asyncsession.go
justinjc: That's definitely one possibility. However, the caller can imitate this behavior by passing in and waiting for the done channel when instantiating this session before calling `Write`, etc, so I feel like this way should provide flexibility.

arnikola: nit: Maybe to resolve this and previous point, `NewSession` could return `(*Session, chan struct{})` to make it a bit more explicit?

asyncsession/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package asyncsession
```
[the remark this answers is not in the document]
justinjc: I'll put it in `stores/m3db/` as we discussed and call it `AsyncSession`.

asyncsession/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+	if s.err != nil {
+		return s.err
+	}
+
+	return s.session.Write(namespace, id, t, value, unit, annotation)
+}
+
+// WriteTagged value to the database for an ID and given tags.
+func (s *Session) WriteTagged(namespace, id ident.ID, tags ident.TagIterator, t time.Time, value float64, unit xtime.Unit, annotation []byte) error {
+	if s.err != nil {
```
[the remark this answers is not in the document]
justinjc: The session gets created with the error `errSessionUninitialized` and gets set to `nil` upon successful initialization or whatever error it failed on. I thought about creating enum states for this session but thought it might be a little over engineering it.

services/m3coordinator/handler/namespace/common.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return nsMap.Metadatas(), value.Version(), nil
+}
+
+// RegisterRoutes registers the namespace routes
+func RegisterRoutes(r *mux.Router, store kv.Store) {
+	logged := logging.WithResponseTimeLogging
+
+	r.HandleFunc(GetURL, logged(NewGetHandler(store)).ServeHTTP).Methods("GET")
```
nikunjgit: Having both GetURL and GetHTTPMethodURL is a little strange. If you go by REST principals, you should just have GetHTTPMethodURL

services/m3coordinator/handler/placement/common.go
```diff
@@ -112,26 +106,19 @@ func ConvertInstancesProto(instancesProto []*placementpb.Instance) ([]placement.
...
 
-	if err != nil {
-		if err == kv.ErrNotFound {
-			return []namespace.Metadata{}, nil
-		}
+	r.HandleFunc(InitURL, logged(NewInitHandler(service)).ServeHTTP).Methods("POST")
 
-		return nil, err
-	}
+	r.HandleFunc(GetURL, logged(NewGetHandler(service)).ServeHTTP).Methods("GET")
```
nikunjgit: same comment as above. Consider only one path for GET   (via GetHTTPMethodURL) and only one path for Delete (DeleteURL)

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -148,8 +142,3 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, re
 	// todo(braskin): implement query execution
 	return nil, errors.New("not implemented")
 }
-
-// SetEngine sets the engine of the handler
-func (h *PromReadHandler) SetEngine(engine *executor.Engine) {
```
nikunjgit: Awesome! :)

services/m3coordinator/main/main.go
```diff
@@ -97,37 +98,24 @@ func main() {
...
 	}
 
-	// Instantiating an M3DB session requires a topology. This function watches (blocks)
-	// for that, so this needs to happen after the potential embedded KV and corresponding
-	// endpoints are set up.
-	session, err := m3dbClient.NewSession()
-	if err != nil {
-		logger.Fatal("unable to create m3db client session", zap.Any("error", err))
-	}
+	session, _ := m3db.NewAsyncSession(m3dbClient)
```
nikunjgit: if you don't fetch from the done chan, wouldn't the go routine block forever and create a resource leak ?

stores/m3db/asyncsession.go
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
nikunjgit: wrong cert

util/util.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
nikunjgit: util/util.go is weird :( maybe util/strings.go ? So that in the future we can put any string related util functions here

services/m3coordinator/handler/namespace/common.go
justinjc: Agreed, we should have a clean interface. I've removed the duplicate paths to the same functionality and now use the DELETE method for delete functionality.

stores/m3db/asyncsession.go
justinjc: Good catch!

services/m3coordinator/main/main.go
justinjc: Going back to the idiomatic (?) way of taking a done `chan` as an argument instead.

services/m3coordinator/handler/namespace/delete.go
```diff
@@ -0,0 +1,117 @@
...
+// THE SOFTWARE.
+
+package namespace
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/gorilla/mux"
```
nikunjgit: import order

services/m3coordinator/handler/placement/delete_test.go
```diff
@@ -0,0 +1,66 @@
...
+package placement
+
+import (
+	"errors"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gorilla/mux"
```
nikunjgit: import order ?

services/m3coordinator/handler/namespace/add.go
```diff
@@ -42,74 +43,74 @@ import (
...
 )
 
 var (
-	errMissingNamespaceName = errors.New("must specify namespace name")
+	errMissingRequiredField = errors.New("all attributes must be set")
 )
 
-// namespaceAddHandler represents a handler for placement add endpoint.
-type namespaceAddHandler AdminHandler
+// addHandler represents a handler for namespace add endpoint.
```
arnikola: Don't think we need comments for unexported methods

services/m3coordinator/handler/namespace/delete.go
```diff
@@ -0,0 +1,117 @@
...
+
+// NewDeleteHandler returns a new instance of handler.
+func NewDeleteHandler(store kv.Store) http.Handler {
+	return &deleteHandler{store: store}
+}
+
+func (h *deleteHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	logger := logging.WithContext(ctx)
+	err := h.delete(mux.Vars(r)[namespaceIDVar])
```
nikunjgit: do you want to do an empty string check here for mux.Vars(r)[namespaceIDVar]

services/m3coordinator/handler/namespace/add.go
```diff
@@ -42,74 +43,74 @@ import (
...
+type addHandler Handler
 
-// NewNamespaceAddHandler returns a new instance of handler.
-func NewNamespaceAddHandler(clusterClient m3clusterClient.Client) http.Handler {
-	return &namespaceAddHandler{
-		clusterClient: clusterClient,
-	}
+// NewAddHandler returns a new instance of handler.
```
arnikola: `NewAddHandler returns a new instance of a namespace add handler`?

services/m3coordinator/handler/namespace/add.go
```diff
@@ -42,74 +43,74 @@ import (
...
 	}
 	defer r.Body.Close()
 
 	addReq := new(admin.NamespaceAddRequest)
 	if err := json.Unmarshal(body, addReq); err != nil {
-		return nil, NewParseError(err, http.StatusBadRequest)
+		return nil, handler.NewParseError(err, http.StatusBadRequest)
+	}
+
+	if util.HasEmptyString(addReq.Name, addReq.RetentionPeriod, addReq.BlockSize, addReq.BufferFuture,
```
arnikola: 👍

services/m3coordinator/handler/namespace/add_test.go
```diff
@@ -54,11 +54,41 @@ func TestNamespaceAddHandler(t *testing.T) {
...
+
+	req = httptest.NewRequest("POST", "/namespace/add", strings.NewReader(jsonInput))
+	require.NotNil(t, req)
+
+	mockKV.EXPECT().Get(M3DBNodeNamespacesKey).Return(nil, kv.ErrNotFound)
+	mockKV.EXPECT().CheckAndSet(M3DBNodeNamespacesKey, gomock.Any(), gomock.Not(nil)).Return(1, nil)
+	addHandler.ServeHTTP(w, req)
+
+	resp = w.Result()
+	body, _ = ioutil.ReadAll(resp.Body)
```
arnikola: nit: worth checking error not nil?

services/m3coordinator/handler/namespace/common.go
```diff
@@ -0,0 +1,76 @@
...
+// Handler represents a generic handler for namespace endpoints.
+type Handler struct {
+	store kv.Store
+}
+
+// Metadata returns the current metadata in the given store and its version
+func Metadata(store kv.Store) ([]namespace.Metadata, int, error) {
+	value, err := store.Get(M3DBNodeNamespacesKey)
+	if err != nil {
+		if err == kv.ErrNotFound {
```
arnikola: nit: Can you put a comment here about why it's ErrNotFound is fine here?

services/m3coordinator/handler/namespace/add.go
```diff
@@ -42,74 +43,74 @@ import (
...
-func (h *namespaceAddHandler) namespaceAdd(ctx context.Context, r *admin.NamespaceAddRequest) (nsproto.Registry, error) {
+func (h *addHandler) add(r *admin.NamespaceAddRequest) (nsproto.Registry, error) {
 	var emptyReg = nsproto.Registry{}
-	kv, err := h.clusterClient.KV()
-	if err != nil {
-		return emptyReg, err
-	}
 
-	currentMetadata, err := currentNamespaceMetadata(kv)
+	currentMetadata, version, err := Metadata(h.store)
```
arnikola: Would it be worth shirtcircuiting here if version = -1? That should never pass the `CheckAndSet` right?

justinjc: I think in Go it is assumed that if a returned error is not nil, the other non-error return values are meaningless.

arnikola: Ah never mind I see how it would work

services/m3coordinator/handler/namespace/common_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockKV := kv.NewMockStore(ctrl)
+	require.NotNil(t, mockKV)
+
+	// Test KV get error
+	mockKV.EXPECT().Get(M3DBNodeNamespacesKey).Return(nil, errors.New("unable to get key"))
+	meta, _, err := Metadata(mockKV)
```
arnikola: nit: check `assert.Equal(t, -1, version)`?

services/m3coordinator/handler/namespace/common_test.go
```diff
@@ -0,0 +1,103 @@
...
+	assert.Nil(t, meta)
+	assert.EqualError(t, err, "unable to get key")
+
+	// Test empty namespace
+	mockKV.EXPECT().Get(M3DBNodeNamespacesKey).Return(nil, kv.ErrNotFound)
+	meta, _, err = Metadata(mockKV)
```
arnikola: nit: check `assert.Equal(t, -1, version)`?

services/m3coordinator/handler/namespace/common_test.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+		},
+	}
+	mockMetaValue := kv.NewMockValue(ctrl)
+	mockMetaValue.EXPECT().Unmarshal(gomock.Not(nil)).Return(nil).SetArg(0, registry)
+	mockMetaValue.EXPECT().Version().Return(0)
+
+	// Test namespaces
+	mockKV.EXPECT().Get(M3DBNodeNamespacesKey).Return(mockMetaValue, nil)
+	meta, _, err = Metadata(mockKV)
```
arnikola: nit: check `assert.Equal(t, 0, version)`?

services/m3coordinator/handler/namespace/delete.go
```diff
@@ -0,0 +1,126 @@
...
+	DeleteURL = fmt.Sprintf("/namespace/{%s}", namespaceIDVar)
+)
+
+var (
+	errNamespaceNotFound = errors.New("unable to find a namespace with specified name")
+
+	errEmptyID = errors.New("must specify namespace ID to delete")
+)
+
+// deleteHandler represents a handler for namespace delete endpoint.
```
arnikola: nit: Don't think we need comments for unexported methods

services/m3coordinator/handler/namespace/get.go
```diff
@@ -18,66 +18,56 @@
...
-
-	// NamespaceGetHTTPMethodURL is the url for the placement get handler (with the GET method).
-	NamespaceGetHTTPMethodURL = "/namespace"
+	// GetURL is the url for the namespace get handler (with the GET method).
+	GetURL = "/namespace"
 )
 
-// namespaceGetHandler represents a handler for placement get endpoint.
-type namespaceGetHandler AdminHandler
+// getHandler represents a handler for namespace get endpoint.
```
arnikola: nit: Don't think we need comments for unexported methods

services/m3coordinator/handler/placement/add.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"go.uber.org/zap"
+)
+
+const (
+	// AddURL is the url for the placement add handler (with the POST method).
+	AddURL = "/placement/add"
+)
+
+// addHandler represents a handler for placement add endpoint.
```
arnikola: nit: Don't think we need comments for unexported methods

services/m3coordinator/handler/placement/delete.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var (
+	// DeleteURL is the url for the placement delete handler (with the DELETE method).
+	DeleteURL = fmt.Sprintf("/placement/{%s}", placementIDVar)
+
+	errEmptyID = errors.New("must specify placement ID to delete")
+)
+
+// deleteHandler represents a handler for placement delete endpoint.
```
arnikola: nit: Don't think we need comments for unexported methods

services/m3coordinator/handler/placement/delete_all.go
```diff
@@ -18,52 +18,38 @@
...
-	// PlacementDeleteHTTPMethodURL is the url for the placement delete handler (with the DELETE method).
-	PlacementDeleteHTTPMethodURL = "/placement"
+	// DeleteAllURL is the url for the handler to delete all placements (with the DELETE method).
+	DeleteAllURL = "/placement"
 )
 
-// PlacementDeleteHandler represents a handler for placement delete endpoint.
-type placementDeleteHandler AdminHandler
+// deleteAllHandler represents a handler for the placement delete all endpoint.
+type deleteAllHandler Handler
```
arnikola: nit: Don't think we need comments for unexported methods

services/m3coordinator/handler/placement/get.go
```diff
@@ -18,46 +18,40 @@
...
-
-	// PlacementGetHTTPMethodURL is the url for the placement get handler (with the GET method).
-	PlacementGetHTTPMethodURL = "/placement"
+	// GetURL is the url for the placement get handler (with the GET method).
+	GetURL = "/placement"
 )
 
-// placementGetHandler represents a handler for placement get endpoint.
-type placementGetHandler AdminHandler
+// getHandler represents a handler for placement get endpoint.
```
arnikola: nit: Don't think we need comments for unexported methods

stores/m3db/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+	return s.session.ShardID(id)
+}
+
+// Close the session
+func (s *AsyncSession) Close() error {
+	if s.err != nil {
+		return s.err
+	}
+
+	return s.session.Close()
```
arnikola: Any chance that you have both an error *and* a session?

Might be worth doing something like this?
```golang
defer func(){
 if s.session != nil {
  s.session.Close()
 }
}()
```

stores/m3db/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+// for given IDs begin failing
+func (s *AsyncSession) ShardID(id ident.ID) (uint32, error) {
+	if s.err != nil {
+		return 0, s.err
+	}
+
+	return s.session.ShardID(id)
+}
+
+// Close the session
```
arnikola: nit: Is go convention something like `Close closes the session`?

arnikola: Similar for other comments in this file

stores/m3db/asyncsession.go
```diff
@@ -0,0 +1,151 @@
...
+// Calls to methods while uninitialized will return an uninitialized error. The done channel
+// is to notify the caller that the session has finished _attempting_ to get initialized.
+type AsyncSession struct {
+	session client.Session
+	done    chan struct{}
+	err     error
+}
+
+// NewAsyncSession returns a new Session
+func NewAsyncSession(c client.Client, done chan struct{}) *AsyncSession {
```
arnikola: Would it be better if this function created the `done chan` itself and returned it, rather than having it passed in?

justinjc: I had changed it to doing it this way, but @nikunjgit commented (and I agree with) that the caller may choose to not receive from the `done chan`, causing a memory leak because the go routine would always hang.

Having the caller make the `done chan` themselves and calling makes it safer in this regard. It also seems like how [Go documentation](https://gobyexample.com/channel-synchronization) does it.

PR #43, opened 2018-04-24
Add native endpoint

benraskin92:
Still need to write some more tests

### DIFF
--- a/services/m3coordinator/handler/prometheus/common.go
+++ b/services/m3coordinator/handler/prometheus/common.go
@@ -1,0 +1,72 @@
+package prometheus
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/m3db/m3coordinator/services/m3coordinator/handler"
+
+	"github.com/golang/snappy"
+)
+
+const (
+	maxTimeout     = time.Minute
+	defaultTimeout = time.Second * 15
+)
+
+// RequestParams are the arguments to a call
+type RequestParams struct {
+	Timeout time.Duration
+}
+
+// ParsePromRequest parses a snappy compressed request from Prometheus
+func ParsePromRequest(r *http.Request) ([]byte, *handler.ParseError) {
+	body := r.Body
+	if r.Body == nil {
+		err := fmt.Errorf("empty request body")
+		return nil, handler.NewParseError(err, http.StatusBadRequest)
+	}
+	defer body.Close()
+	compressed, err := ioutil.ReadAll(body)
+
+	if err != nil {
+		return nil, handler.NewParseError(err, http.StatusInternalServerError)
+	}
+
+	if len(compressed) == 0 {
+		return nil, handler.NewParseError(fmt.Errorf("empty request body"), http.StatusBadRequest)
+	}
+
+	reqBuf, err := snappy.Decode(nil, compressed)
+	if err != nil {
+		return nil, handler.NewParseError(err, http.StatusBadRequest)
+	}
+
+	return reqBuf, nil
+}
+
+// ParseRequestParams parses the input request parameters and provides useful defaults
+func ParseRequestParams(r *http.Request) (*RequestParams, error) {
+	var params RequestParams
+	timeout := r.Header.Get("timeout")
+	if timeout != "" {
+		duration, err := time.ParseDuration(timeout)
+		if err != nil {
+			return nil, fmt.Errorf("%s: invalid 'timeout': %v", handler.ErrInvalidParams, err)
+
+		}
+
+		if duration > maxTimeout {
+			return nil, fmt.Errorf("%s: invalid 'timeout': greater than %v", handler.ErrInvalidParams, maxTimeout)
+
+		}
+
+		params.Timeout = duration
+	} else {
+		params.Timeout = defaultTimeout
+	}
+
+	return &params, nil
+}
--- a/services/m3coordinator/handler/prometheus/common.go
+++ b/services/m3coordinator/handler/prometheus/common.go
@@ -55,12 +55,10 @@ func ParseRequestParams(r *http.Request) (*RequestParams, error) {
 		duration, err := time.ParseDuration(timeout)
 		if err != nil {
 			return nil, fmt.Errorf("%s: invalid 'timeout': %v", handler.ErrInvalidParams, err)
-
 		}
 
 		if duration > maxTimeout {
 			return nil, fmt.Errorf("%s: invalid 'timeout': greater than %v", handler.ErrInvalidParams, maxTimeout)
-
 		}
 
 		params.Timeout = duration
--- a/services/m3coordinator/handler/prometheus/common.go
+++ b/services/m3coordinator/handler/prometheus/common.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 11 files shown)


services/m3coordinator/handler/prometheus/common.go
```diff
@@ -0,0 +1,72 @@
...
+	"net/http"
+	"time"
+
+	"github.com/m3db/m3coordinator/services/m3coordinator/handler"
+
+	"github.com/golang/snappy"
+)
+
+const (
+	maxTimeout     = time.Minute
```
arnikola: Looks like these are shared everywhere; maybe use a constants file and/or get it from configs?

services/m3coordinator/handler/prometheus/common.go
```diff
@@ -0,0 +1,72 @@
...
+
+// ParseRequestParams parses the input request parameters and provides useful defaults
+func ParseRequestParams(r *http.Request) (*RequestParams, error) {
+	var params RequestParams
+	timeout := r.Header.Get("timeout")
+	if timeout != "" {
+		duration, err := time.ParseDuration(timeout)
+		if err != nil {
+			return nil, fmt.Errorf("%s: invalid 'timeout': %v", handler.ErrInvalidParams, err)
+
```
arnikola: extra line

services/m3coordinator/handler/prometheus/common.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		if duration > maxTimeout {
+			return nil, fmt.Errorf("%s: invalid 'timeout': greater than %v", handler.ErrInvalidParams, maxTimeout)
+
```
arnikola: extra line

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,133 @@
...
+	PromReadURL = "/api/v1/prom/native/read"
+)
+
+// PromReadHandler represents a handler for prometheus read endpoint.
+type PromReadHandler struct {
+	engine *executor.Engine
+}
+
+// ReadResponse is the response that gets returned to the user
+type ReadResponse struct {
```
arnikola: Can be private?

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/golang/snappy"
+	"go.uber.org/zap"
+)
+
+const (
+	// PromReadURL is the url for native prom read handler
+	PromReadURL = "/api/v1/prom/native/read"
+)
+
+// PromReadHandler represents a handler for prometheus read endpoint.
```
arnikola: Isn't there already an existing PromReadHandler? Might be good to change the names so it's not just a namespace distinction.

(also can possibly be private?)

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, err
+	}
+
+	// todo(braskin): implement query execution
+	return nil, nil
+}
+
+// SetEngine sets the engine of the handler
+func (h *PromReadHandler) SetEngine(engine *executor.Engine) {
+	h.engine = engine
```
arnikola: Would this have to stop the existing engine?

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,133 @@
...
+	Results []*ts.Series `json:"results,omitempty"`
+}
+
+// NewPromReadHandler returns a new instance of handler.
+func NewPromReadHandler(engine *executor.Engine) http.Handler {
+	return &PromReadHandler{engine: engine}
+}
+
+func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	params, err := prometheus.ParseRequestParams(r)
```
arnikola: Any chance h.engine is nil here if it hasn't been lazy loaded yet?

justinjc: Right, you should put in a snippet like [this](https://github.com/m3db/m3coordinator/blob/ef0c47e9961f2ec2e3b066680caa7c611c3d6472/services/m3coordinator/handler/read.go#L57-L61), since there might not be an engine at this point.

services/m3coordinator/handler/prometheus/native/read.go
justinjc: Good point - I'll include this in my PR. I'm thinking of allowing it to be set only if it is nil.

services/m3coordinator/handler/prometheus/native/read.go
benraskin92: We'll have a `remote` and a `native` package so it'd be `remote.PromReadHandler` and `native.PromReadHandler`

services/m3coordinator/handler/prometheus/native/read.go
benraskin92: I think I'll need access to it in my next PR

services/m3coordinator/handler/prometheus/common.go
benraskin92: Forgot to remove these from the other file so this is the only place they're used.

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// PromReadHandler represents a handler for prometheus read endpoint.
+type PromReadHandler struct {
+	engine *executor.Engine
+}
+
+// ReadResponse is the response that gets returned to the user
+type ReadResponse struct {
+	Results []*ts.Series `json:"results,omitempty"`
```
nikunjgit: just have []ts.Series ?

services/m3coordinator/handler/prometheus/native/read.go
nikunjgit: I think you need to be public for json.Marshal

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,141 @@
...
+
+	data, err := json.Marshal(resp)
+	if err != nil {
+		logger.Error("unable to marshal read results to json", zap.Any("error", err))
+		handler.Error(w, err, http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Content-Encoding", "snappy")
```
nikunjgit: we shouldn't do snappy encoding here. Probably look at how statsdex is sending data to grafana.

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,141 @@
...
+	compressed := snappy.Encode(nil, data)
+	if _, err := w.Write(compressed); err != nil {
+		logger.Error("unable to encode read results to snappy", zap.Any("err", err))
+		handler.Error(w, err, http.StatusInternalServerError)
+		return
+	}
+}
+
+func (h *PromReadHandler) parseRequest(r *http.Request) (string, *handler.ParseError) {
+	reqBuf, err := prometheus.ParsePromRequest(r)
```
nikunjgit: I think you need a GET method which accepts a query string

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// todo(braskin): implement query execution
+	return nil, errors.New("not implemented")
+}
+
+// SetEngine sets the engine of the handler
+func (h *PromReadHandler) SetEngine(engine *executor.Engine) {
```
nikunjgit: why do you need this ?

services/m3coordinator/httpd/handler.go
```diff
@@ -140,7 +147,8 @@ func withResponseTimeLogging(next http.Handler) http.Handler {
 
 // LoadLazyHandlers initializes LazyHandlers post-M3DB setup
 func (h *Handler) LoadLazyHandlers(storage storage.Storage, engine *executor.Engine) {
```
nikunjgit: why are we optimizing for this usecase ? We probably want to wait until M3DB connections are actually established before starting the query service

services/m3coordinator/handler/prometheus/native/read.go
benraskin92: cc @justinjc

services/m3coordinator/httpd/handler.go
benraskin92: cc @justinjc

services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if _, err := w.Write(data); err != nil {
+		logger.Error("unable to write results", zap.Any("err", err))
+		handler.Error(w, err, http.StatusInternalServerError)
+		return
+	}
+}
+
+func (h *PromReadHandler) parseRequest(r *http.Request) (string, *handler.ParseError) {
+	targetQueries := r.URL.Query()[targetQuery]
```
nikunjgit: targetQueries, ok ?

PR #45, opened 2018-04-27
Stop skipping test

benraskin92:


benraskin92:
I think you need to remove those changes to the `.ci` folder- I updated them for the migration to Semaphore. Looks like the build is failing because of this.

tsdb/remote/server_test.go
```diff
@@ -356,39 +355,51 @@ func TestRoundRobinClientRpc(t *testing.T) {
...
-
-		// Write called on hostß
-		checkWrite(ctx, t, client, write)
+	// Host ordering is not always deterministic; retry several times to ensure at least one
+	// call is made to both hosts. Giving 10 attempts per host should remove flakiness while guaranteeing
+	// round robin behaviour
+	attempts := 20
+
+	hitHost, hitErrHost := false, false
+	for i := 0; i < attempts; i++ {
```
nikunjgit: so round robin should cycle through connections on every request and seems like that's the case if you look here: https://github.com/grpc/grpc-go/blob/e6549e636da11b0ca0379af8bdb7d6d22ea80893/balancer/roundrobin/roundrobin.go. The fact our test is failing, does that point to an issue in the code itself in the way we initialize the connections ?

arnikola: Talked offline; really difficult to repro and potentially not useful since we want to move to a dynamic round robin implementation soon.

PR #46, opened 2018-04-28
Update glide deps to use to the same versions m3db/m3db

justinjc:
**NOTE:** The new index currently does not support negate queries, so we may want to hold off on landing this until that functionality is there.


generated/proto/admin/namespace.proto
```diff
@@ -16,10 +16,10 @@ message NamespaceAddRequest {
...
   bool block_data_expiry = 6;
   string block_data_expiry_period = 7;
-  bool needs_bootstrap = 8;
-  bool needs_fileset_cleanup = 9;
-  bool needs_flush = 10;
-  bool needs_repair = 11;
+  bool bootstrap_enabled = 8;
+  bool cleanup_enabled = 9;
+  bool flush_enabled = 10;
+  bool repair_enabled = 11;
```
justinjc: These variable renames are not strictly required, but it's nice to align these with the method names from m3db.

storage/index.go
```diff
@@ -88,12 +99,23 @@ func MatchersToFilters(matchers models.Matchers) []segment.Filter {
...
+				return nil, err
+			}
+		} else {
+			query = idx.NewTermQuery([]byte(matcher.Name), []byte(matcher.Value))
+		}
+
+		if negate {
+			// TODO: implement this when idx supports this
+			// query = idx.NewNegateQuery(query)
+			return nil, xerrors.ErrNotImplemented
```
justinjc: Note that negation is not yet supported (but should be shortly). The inside scoop from @jeromefroe is that the interface should be something like the commented line above.

storage/local/storage.go
```diff
@@ -115,9 +115,13 @@ func (s *localStorage) FetchTags(ctx context.Context, query *storage.FetchQuery,
...
 
-	m3query := storage.FetchQueryToM3Query(query)
+	m3query, err := storage.FetchQueryToM3Query(query)
+	if err != nil {
+		return nil, err
+	}
+
 	opts := storage.FetchOptionsToM3Options(options, query)
-	results, err := s.session.FetchTaggedIDs(m3query, opts)
+	results, err := s.session.FetchTaggedIDs(ident.StringID(s.namespace), m3query, opts)
```
justinjc: Unsure whether this is correct here.

benraskin92: maybe pull `ident.StringID(s.namespace)` into a var

i.e. `name := ident.StringID(s.namespace)`

storage/index.go
```diff
@@ -42,9 +43,11 @@ func FromM3IdentToMetric(identNamespace, identID ident.ID, iterTags ident.Tags)
...
-func FromIdentTagsToTags(identTags ident.Tags) models.Tags {
-	tags := make(models.Tags, len(identTags))
-	for _, identTag := range identTags {
+func FromIdentTagsToTags(identTags ident.TagIterator) models.Tags {
+	tags := make(models.Tags, identTags.Remaining())
+	for identTags.Next() {
+		identTag := identTags.Current()
+
 		tags[identTag.Name.String()] = identTag.Value.String()
 		identTag.Finalize()
```
arnikola: I think since we're using a tagIterator, I believe we can finalize the entire iterator in a defer or after the loop

storage/index.go
```diff
@@ -61,23 +64,31 @@ func FetchOptionsToM3Options(fetchOptions *FetchOptions, fetchQuery *FetchQuery)
...
+	}
+
+	query, err := idx.NewConjunctionQuery(queries...)
+	if err != nil {
+		return index.Query{}, err
 	}
-	indexQuery.Query = segQuery
 
-	return indexQuery
+	var indexQuery index.Query
```
arnikola:
```
	return index.Query{
		Query: segQuery,
	},nil
```

storage/index.go
prateek: +1, also the tag iterator's `Finalize()` method has been renamed `Close()`

storage/index.go
```diff
@@ -42,9 +43,11 @@ func FromM3IdentToMetric(identNamespace, identID ident.ID, iterTags ident.Tags)
 }
 
 // FromIdentTagsToTags converts ident tags to coordinator tags
-func FromIdentTagsToTags(identTags ident.Tags) models.Tags {
-	tags := make(models.Tags, len(identTags))
-	for _, identTag := range identTags {
+func FromIdentTagsToTags(identTags ident.TagIterator) models.Tags {
+	tags := make(models.Tags, identTags.Remaining())
+	for identTags.Next() {
```
prateek: Unrelated to your changes, this was missing in the original code too: Need to check `identTags.Err()` after the loop and propagate error up. Can address now/open an issue as you see fit.

storage/index.go
```diff
@@ -61,23 +64,31 @@ func FetchOptionsToM3Options(fetchOptions *FetchOptions, fetchQuery *FetchQuery)
...
 
-	return indexQuery
+	var indexQuery index.Query
+	indexQuery.Query = query
+
+	return indexQuery, nil
 }
 
-// MatchersToFilters converts matchers to M3 filters
-func MatchersToFilters(matchers models.Matchers) []segment.Filter {
```
arnikola: I believe you can delete segment.Filter since this is the only place it's used

storage/local/storage.go
arnikola: That seems fine, but would probably want to also `defer id.Finalize()` since ident.StringId pools

storage/index.go
```diff
@@ -61,23 +64,31 @@ func FetchOptionsToM3Options(fetchOptions *FetchOptions, fetchQuery *FetchQuery)
...
+// MatchersToQueries converts matchers to M3 filters
+func MatchersToQueries(matchers models.Matchers) ([]idx.Query, error) {
 	var (
-		filters []segment.Filter
+		queries []idx.Query
```
arnikola: nit `queries := make([] idx.Query, 0, len(matchers))`

storage/index.go
```diff
@@ -88,12 +99,23 @@ func MatchersToFilters(matchers models.Matchers) []segment.Filter {
 			regexp = true
 		}
 
-		filters = append(filters, segment.Filter{
-			FieldName:        []byte(matcher.Name),
-			FieldValueFilter: []byte(matcher.Value),
-			Negate:           negate,
-			Regexp:           regexp,
-		})
+		if regexp {
```
arnikola: Wouldn't this need to check all four matchers in one conditional?

justinjc: Not entirely sure what you mean by this. Since the interface for `negate` queries is something like `idx.negate(query)`, I think it's fine to first construct a Regexp or Term query and then negate it after in a later conditional (see my comment below).

arnikola: Yeah didn't explain well, but wouldn't we have to do something like:

```golang
if regexp {
 if negate {
  query, _ = idx.NewNegateRegexpQuery([]byte(matcher.Name), []byte(matcher.Value))
 } else {
  query, _ = idx.NewRegexpQuery([]byte(matcher.Name), []byte(matcher.Value))
 }
} else {
 if negate {
  query, _ = idx.NewNegateQuery([]byte(matcher.Name), []byte(matcher.Value))
 } else {
  query = idx.NewTermQuery([]byte(matcher.Name), []byte(matcher.Value))
 }
}
```

arnikola: Actually reading over the projected interface it's fine

storage/index.go
```diff
@@ -21,34 +21,47 @@
...
 		Tags:      tags,
-	}
+	}, nil
 }
 
 // FromIdentTagsToTags converts ident tags to coordinator tags
-func FromIdentTagsToTags(identTags ident.Tags) models.Tags {
-	tags := make(models.Tags, len(identTags))
-	for _, identTag := range identTags {
+func FromIdentTagsToTags(identTags ident.TagIterator) (models.Tags, error) {
```
nikunjgit: FromIdentTagIteratorToTags

storage/index.go
```diff
@@ -21,34 +21,47 @@
...
+
+	for identTags.Next() {
+		identTag := identTags.Current()
+
 		tags[identTag.Name.String()] = identTag.Value.String()
-		identTag.Finalize()
 	}
-	return tags
+
+	defer identTags.Close()
```
nikunjgit: i'd say not close the iterator here since the caller wouldn't have any idea.

storage/index.go
nikunjgit:
```
func matchToQuery(matcher *models.Matcher) (idx.Query, error) {
	switch matcher.Type {
	case models.MatchRegexp:
		return idx.NewRegexpQuery([]byte(matcher.Name), []byte(matcher.Value))

	case models.MatchEqual:
		return idx.NewTermQuery([]byte(matcher.Name), []byte(matcher.Value)), nil

	default:
		return idx.Query{}, fmt.Errorf("unsupported query type %v", matcher)

	}
}
```

nikunjgit: caller can be:

```
func FetchQueryToM3Query(fetchQuery *FetchQuery) (index.Query, error) {
	matchers := fetchQuery.TagMatchers
	idxQueries := make([]idx.Query, len(matchers))
	var err error
	for i, matcher := range matchers {
		idxQueries[i], err = matchToQuery(matcher)
		if err != nil {
			return index.Query{}, err
		}
	}

	q, err := idx.NewConjunctionQuery(idxQueries...)
	return index.Query{Query: q}, err
}

```

generated/proto/admin/namespace.proto
nikunjgit: sg

storage/local/storage.go
```diff
@@ -115,9 +115,16 @@ func (s *localStorage) FetchTags(ctx context.Context, query *storage.FetchQuery,
 	default:
 	}
 
-	m3query := storage.FetchQueryToM3Query(query)
+	m3query, err := storage.FetchQueryToM3Query(query)
+	if err != nil {
+		return nil, err
+	}
+
+	id := ident.StringID(s.namespace)
```
nikunjgit: instead of storing namespace as string, you can just store it as ident.ID ? That will avoid a lot of these conversions

justinjc: Spoke offline - we'll include this change in a separate diff that will include a Close method which will be the logical place to finalize the ident.ID.

PR #48, opened 2018-05-07
Add Blocks docs

benraskin92:


docs/architecture/blocks.md
```diff
@@ -0,0 +1,63 @@
...
+┌───────────────────────────┐ │      │      │      │    │      │      │      │   │      │      │      │
+│       name:sign_up        │ │      │      │      │    │      │      │      │   │      │      │      │
+│  city_id:toronto os:ios   │ │  2   │  5   │  10  │    │  2   │  5   │  10  │   │  2   │  5   │  10  │
+└───────────────────────────┘ │      │      │      │    │      │      │      │   │      │      │      │
+                              └──────┴──────┴──────┘    └──────┴──────┴──────┘   └──────┴──────┴──────┘
+```
+
+## M3DB => M3Coordinator Blocks
+
+M3DB will return to M3Coordinator `SeriesBlocks` that contain a list of `SeriesIterators` for a given timeseries per namespace. We would convert that into a `ConslidatedBlockIterators` depending on the consolidation chosen. Each underlying `ConsolidatedBlockIterator` would have the same time duration. This means that the last block can span outside the `EndTime`. We would rely on the user of these blocks to perform the truncation.
```
nikunjgit: Probably remove the last line about truncation. I think SeriesIterators should be already handing that so we shouldn't need to rely on the client.

docs/architecture/blocks.md
```diff
@@ -0,0 +1,63 @@
...
+
+From `ConslidatedBlockIterators` per series, we will construct a `MultiSeriesBlock` that will contain a slice of `ConslidatedBlockIterators` for a given block as well as a `Start` and `End`. `MultiSeriesBlock` aligns the different series to their LCM.
```
nikunjgit: elaborate a little bit on alignment

PR #53, opened 2018-05-10
Fetch working with count using the block interface

nikunjgit:


functions/count.go
```diff
@@ -54,3 +55,27 @@ type CountNode struct {
...
+// Process the block
+func (c *CountNode) Process(ID parser.NodeID, block storage.Block) error {
+	builder, err := c.controller.BlockBuilder(block.Meta())
+	if err != nil {
+		return err
+	}
+
+	stepIter := block.StepIter()
+	index := 0
+	for stepIter.Next() {
```
justinjc: `for index := 0; stepIter.Next(); index++ {`

functions/fetch.go
```diff
@@ -60,11 +61,31 @@ func (o FetchOp) String() string {
...
 // Execute runs the fetch node operation
 func (n *FetchNode) Execute(ctx context.Context) error {
-	return fmt.Errorf("not implemented")
+	startTime := n.now.Add(-1 * n.op.Offset)
+	endTime := startTime.Add(n.op.Range)
+	blockResult, err := n.storage.FetchBlocks(ctx, &storage.FetchQuery{
+		Start:       startTime,
+		End:         endTime,
+		TagMatchers: n.op.Matchers,
+	}, &storage.FetchOptions{})
```
justinjc: Would we ever want to pass different arguments to this apart from using the default struct?

functions/fetch.go
```diff
@@ -60,11 +61,31 @@ func (o FetchOp) String() string {
...
+		return err
+	}
+
+	for _, block := range blockResult.Blocks {
+		err := n.controller.Process(block)
+		// Fail on first error
+		if err != nil {
+			return err
+		}
+
```
justinjc: Nit: remove extra line.

functions/fetch.go
```diff
@@ -60,11 +61,31 @@ func (o FetchOp) String() string {
...
+		TagMatchers: n.op.Matchers,
+	}, &storage.FetchOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, block := range blockResult.Blocks {
+		err := n.controller.Process(block)
+		// Fail on first error
+		if err != nil {
```
justinjc: Perhaps `if err := n.controller.Process(block); err != nil {`

functions/count.go
nikunjgit: oops :-)

functions/fetch.go
nikunjgit: ofcourse, some things would be ID limit, concurrency, etc.

executor/transform/controller.go
```diff
@@ -21,16 +21,43 @@
...
+}
+
+// BlockBuilder returns a BlockBuilder instance with associated metadata
+// nolint: unparam
+func (t *Controller) BlockBuilder(blockMeta storage.BlockMetadata) (BlockBuilder, error) {
+	return nil, fmt.Errorf("not implemented")
+}
+
+// BlockBuilder builds a new block
+type BlockBuilder interface {
```
benraskin92: This is like the resulting `block` after `process` is called?

nikunjgit: Yeah, I use that in count, so you can see how we might be able to use it

storage/block.go
```diff
@@ -0,0 +1,58 @@
...
+// StepMeta is metadata data for a single time step
+type StepMeta struct {
+}
+
+// Bounds are the time bounds
+// nolint: structcheck, megacheck
+type Bounds struct {
+	start    time.Time
+	end      time.Time
+	stepSize time.Duration
```
benraskin92: Any reason we should also include `numSteps`? I know it can be derived, but just trying to think of any use cases.

justinjc (approved):
LGTM

storage/block.go
```diff
@@ -0,0 +1,58 @@
...
+	Meta() BlockMetadata
+	StepIter() StepIter
+	SeriesIter() SeriesIter
+	SeriesMeta() []SeriesMeta
+	StepMeta() []StepMeta
+}
+
+// SeriesMeta is metadata data for the series
+type SeriesMeta struct {
+	Tags models.Tags
```
benraskin92: These are the unique tags, right? i.e. `SeriesMeta.Tags` + `BlockMetaData.Tags` = `Series.Tags` (or something like that).

storage/block.go
nikunjgit: we can add later if needed ?

storage/local/storage.go
```diff
@@ -171,6 +172,11 @@ func (s *localStorage) Type() storage.Type {
 	return storage.TypeLocalDC
 }
 
+func (s *localStorage) FetchBlocks(
+	ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (storage.BlockResult, error) {
+	return storage.BlockResult{}, fmt.Errorf("not implemented")
```
benraskin92: nit: maybe just `errors.New("not implemented")`? (in other places as well)

storage/block.go
nikunjgit: actually I was thinking that maybe this can have all tags and then blocks will have common tags. It's an implementation detail so maybe we can look at it later ?

PR #54, opened 2018-05-10
Make coordinator runnable on docker config

justinjc:
Includes nits suggested by @arnikola from #41 after it got merged.

### DIFF
--- a/docker/coordinator.yml
+++ b/docker/coordinator.yml
@@ -1,12 +1,12 @@
-backend: m3db
-backendListen: 0.0.0.0:9000
-
 client:
   config:
     static:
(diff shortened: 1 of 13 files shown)


docker/coordinator.yml
```diff
@@ -1,12 +1,12 @@
-backend: m3db
-backendListen: 0.0.0.0:9000
-
 client:
   config:
     static:
       listenAddress: 0.0.0.0:9000
       topology:
           shards: 2
+          hosts:
```
justinjc: The config needs to specify at least one host here. Not sure what to put though.

